PR #3, opened 2023-05-11
feat: add x/exp/ordered for generics-based min, max, and clamp utilities

meowgorithm:
This simple feature adds `min`, `max`, and `clamp` functions for `constraints.Ordered` types.

Suggestions welcome.


exp/clamp/clamp.go
```diff
@@ -0,0 +1,26 @@
+package clamp
```
caarlos0, line 1: maybe `x/exp/ordered` instead? (as currently all functions use `constraints.Ordered`) 🤔

caarlos0 (approved):
lgtm, just missing making the funcs public

exp/clamp/clamp.go
meowgorithm: Ah yeah, that's better. Will adjust.

PR #4, opened 2023-05-12
feat: add log mr slog

aymanbagabas:
add charm log with slog support
this depends on golang.org/x/exp/slog

this also changes the module path from `github.com/charmbracelet/log` to `github.com/charmbracelet/x/exp/log`


caarlos0 (approved):
might be painful to have both in sync, but other than that LGTM

caarlos0:
fwiw slog apparently will be in go1.21 👀

aymanbagabas:
[Log](https://github.com/charmbracelet/log) now supports `log/slog`

PR #9, opened 2023-07-07
feat: generic slice 'take' function

meowgorithm:
`Take` returns a sub-slice containing at most _N_ items from the head of a slice. I know Go’s not tail-call optimized, but I can still see this being handy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Currently the following packages are available:
 
 * [`higherorder`](./exp/higherorder): generic higher order functions
 * [`ordered`](./exp/ordered): generic `min`, `max`, and `clamp` functions for ordered types
+* [`slice`](./exp/slice): generic slice utilities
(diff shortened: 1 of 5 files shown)


maaslalani:
I wonder if calling this `Head` makes sense, which would allow us to also introduce `Tail` to "take" the last `N` elements of an array?

meowgorithm:
I think `Head` and `Tail` make a lot of logical sense, though in FP `head` usually returns a single element and `tail` returns a list of everything except the first item, so I'd be concerned about confusion there. `Take` is also a known FP term for getting the first N items.

Perhaps `TakeHead` could be an alias for `Take` and then `TakeTail` could take N item from the end.

maaslalani:
> Take is also a known FP term for getting the first N items.

Ah, cool. I didn't know about this. Then this makes sense to me for sure!

maaslalani (approved):

caarlos0 (approved):

PR #12, opened 2023-10-03
feat: add generic reverse and delete functions for slices

meowgorithm:
Just a couple o’ funcs I’d love to stop writing over and over.

### DIFF
--- a/exp/slice/slice.go
+++ b/exp/slice/slice.go
@@ -8,3 +8,25 @@ func Take[A any](slice []A, n int) []A {
 	}
 	return slice[:n]
 }
+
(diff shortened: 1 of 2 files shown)


caarlos0:
I think you can do all that with go1.21 `slices` package...

caarlos0 (approved):

meowgorithm:
> I think you can do all that with go1.21 `slices` package...
> 
> not sure if the idea is to have this here too for older go versions, but thought I'd point it out anyway :P

Welp TIL! Actually had just not realized this is all in stdlib now. Closing accordingly!

PR #13, opened 2023-10-10
feat: add ordered.First, tests, use stdlib cmp

caarlos0:
I have written hundreds of `firstNonEmpty` functions over the years, guess this is a good place for it.

Also using stdlib `cmp` instead of `/x/exp/constraints`.

And 100% test coverage :)


meowgorithm (approved):
❤️

PR #14, opened 2023-10-25
feat: join strings with a conjunction in a handful of languages

meowgorithm:
At last, join strings in the way they would be read. For example, `[]string{"cheese", "beans", "jalapeños"}` can become `"cheese, beans, and jalapeños"`, with a flick of the wrist.

```go
import str "gitub.com/charmbracelet/x/exp/strings"

x := []string{"eins", "zwei", "drei"}
s := str.SpokenLanguageJoin(x, str.DE)
fmt.Println(s) // eins, zwei und drei
```

English users can also opt into to the Oxford Comma with the hot `EnglishJoin` function:

```go
x := []string{"meow", "purr", "raow"}
s := str.EnglishJoin(x, true)
fmt.Println(s) // meow, purr, and raow
```

The following languages are supported:

* German (DE)
* Danish (DK)
* English (EN)
* Spanish (ES)
* French (FR)
* Italian (IT)
* Norwegian ([NO](https://www.bram.us/2022/01/11/yaml-the-norway-problem/))
* Portuguese (PT)
* Swedish (SE)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,24 +16,25 @@ into other repositories.
 
 Currently the following packages are available:
 
-* [`higherorder`](./exp/higherorder): generic higher order functions
-* [`ordered`](./exp/ordered): generic `min`, `max`, and `clamp` functions for ordered types
-* [`slice`](./exp/slice): generic slice utilities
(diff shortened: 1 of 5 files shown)


caarlos0:
LOVE THIS!

PR #17, opened 2023-12-10
feat(editor): Ability to Pass Optional Line Number Argument to `Cmd`

h0rv:
This is needed for a feature I want to implement in Glow: https://github.com/charmbracelet/glow/issues/547.


h0rv:
Hi @meowgorithm, would you mind reviewing this PR when you get a change? Thanks!

caarlos0:
maybe instead we pass a `lineNumber` arg, and handle it here? so it can be used in other projects as well

h0rv:
Ok I have logic in a local Charm PR to add this. Right now, I know for certain that `nano`, `vi`, `vim`, and `nvim` support this feature.

And how about leaving the current change being able to pass in arbitrary args, and then creating a new function a new function: `openEditorAtLine`?

caarlos0:
I think we can leave it out unless we find a good use case for it 🤔

h0rv:
Updated to only include `lineNumber`

caarlos0:
see  #26

PR #18, opened 2023-12-11
made cmp import work correctly

Static-Flow:
Was trying to use the new `huh` library but it imports this lib which in turn attempts to import `cmp` but from the GOROOT instead of GOPATH. This fixes that

### DIFF
--- a/exp/ordered/ordered.go
+++ b/exp/ordered/ordered.go
@@ -1,6 +1,6 @@
 package ordered
 
-import "cmp"
+import "github.com/google/go-cmp/cmp"
 
 // Max returns the smaller of a and b.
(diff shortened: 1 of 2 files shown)


caarlos0:
that's not correct, we're in fact using the stdlib cmp package 

https://pkg.go.dev/cmp@go1.21.5

if you're having issues, you likely need a more recent go version 🤔

Static-Flow:
👍  Can we put that in the huh readme then that it only works with specific Go versions? Sadly there's nothing in the import errors that would lead one to know that is the issue.

caarlos0:
ah, this should do the trick https://github.com/charmbracelet/huh/pull/42

thanks for the heads up

PR #27, opened 2023-12-20
feat: add maps.SortedKeys

caarlos0:
added maps.SortedKeys, can add more methods in the future.

This one simply combines `x/maps.Keys` and `x/slices.Sort`, and is merely a convenience.


exp/maps/maps.go
```diff
@@ -0,0 +1,15 @@
+package maps
+
+import (
+	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
```
aymanbagabas, line 6:
```suggestion
	"slices"
```

exp/maps/maps.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+// SortedKeys returns the keys of the map m.
+// The keys will be sorted.
+func SortedKeys[M ~map[K]V, K constraints.Ordered, V any](m M) []K {
```
aymanbagabas, line 11:
```suggestion
func SortedKeys[M ~map[K]V, K cmp.Ordered, V any](m M) []K {
```

exp/maps/maps.go
```diff
@@ -0,0 +1,15 @@
+package maps
+
+import (
+	"golang.org/x/exp/constraints"
```
aymanbagabas, line 4:
```suggestion
	"cmp"
```

exp/maps/maps_test.go
```diff
@@ -0,0 +1,20 @@
+package maps
+
+import (
+	"testing"
+
+	"golang.org/x/exp/slices"
```
aymanbagabas, line 6:
```suggestion
	"slices"
```

aymanbagabas (commented):
We could use standard library `slices` and `cmp` packages to limit the use of `x/exp`

exp/maps/maps.go
caarlos0: will require a highger go version... which is fine I think

ccoVeille (approved):

PR #28, opened 2024-01-09
feat(term): add term/conpty

aymanbagabas:
Add support for windows conpty https://learn.microsoft.com/en-us/windows/console/creating-a-pseudoconsole-session

### DIFF
--- a/term/conpty/conpty_windows.go
+++ b/term/conpty/conpty_windows.go
@@ -96,9 +96,9 @@ func New(w int, h int, flags int) (c *ConPty, err error) {
 	return
 }
 
-// Handle returns the ConPty handle.
-func (p *ConPty) Handle() windows.Handle {
-	return *p.hpc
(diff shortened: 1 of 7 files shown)


term/conpty/conpty_windows.go
```diff
@@ -0,0 +1,271 @@
+package conpty
```
caarlos0, line 1: Should we add `go:build windows`?

caarlos0 (commented):
Lgtm, probably good to re-run the scripts that gen the CI scripts too

term/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/charmbracelet/x/term
+
+go 1.21
```
caarlos0, line 3: can we lower this?

aymanbagabas: done

caarlos0 (approved):
NOICE

PR #32, opened 2024-01-12
feat(term): add ansi sequences support

aymanbagabas:
This package define common ansi sequences and control codes used in a terminal session.

For example, `style.String` to style a string using the SGR escape sequence. `style.String("Foobar", style.Bold, style.ForegrondColor(style.Red))` will produce a sequence that wraps the word "Foobar" in two SGR sequences to display the word in bold and red foreground color, and then reset SGR styles.

Another example is using the `mode` package to enable/disable/request various terminal modes. E.g. to enable alternate screen buffer mode `fmt.Fprint(os.Stderr, mode.EnableAltScreenBuffer)`.


maaslalani (approved):
This is so awesome! @aymanbagabas ❤️

caarlos0 (approved):

PR #35, opened 2024-02-01
feat(term): termios package

caarlos0:
This will help us remove the dep on u-root/u-root from /ssh, and also adds support to netbsd.

The api is different, and it does not use reflection.


aymanbagabas (commented):
Noice! Does this build on all BSDs? Should we support `dragonfly`? Would be nice to have a workflow to build for all these OSes

caarlos0:
Assuming dragonfly uses the same syscalls, its possible yeah.

agreed, wouldn't hurt to `go build` for all `GOOS/GOARCH` to see if it at least compiles.

exp/term/termios/termios.go
```diff
@@ -0,0 +1,161 @@
...
+	"ocrnl":  {O, syscall.OCRNL},
+	"onocr":  {O, syscall.ONOCR},
+	"onlret": {O, syscall.ONLRET},
+
+	"cs7":    {C, syscall.CS7},
+	"cs8":    {C, syscall.CS8},
+	"parenb": {C, syscall.PARENB},
+	"parodd": {C, syscall.PARODD},
+
+	// XXX: not available on some OSs
```
aymanbagabas, line 156: We can define these for the OSs that support them in a `init`

exp/term/termios/termios.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	I = iota // Input control
+	O        // Output control
+	C        // Control
+	L        // Line control
+)
+
+// https://www.man7.org/linux/man-pages/man3/termios.3.html
+var allCcOpts = map[string]int{
+	"intr":    syscall.VINTR,
```
aymanbagabas, line 96: Should we define the strings as consts?

caarlos0: was thinking about it too...

PR #49, opened 2024-03-11
Term: update ANSI parser

aymanbagabas:
- Tidy up and improve `vtparser` and move it to `ansi.Parser`
- Implement `ansi.Strip` and `ansi.StringWidth` to stip ANSI sequences from a string and count the terminal cells width of a string while ignoring ANSI sequences and accounting for wide characters.


maaslalani (approved):

maaslalani:
After this is merged we can probably swap out the implementation of `lipgloss.Width` for `ansi.StringWidth` if they are equivalent:


https://github.com/charmbracelet/lipgloss/blob/439c06fae64d2f53261b692fcfcbe464d8e18d89/size.go#L15

```go
// Width returns the cell width of characters in the string. ANSI sequences are
// ignored and characters wider than one cell (such as Chinese characters and
// emojis) are appropriately measured.
//
// You should use this instead of len(string) len([]rune(string) as neither
// will give you accurate results.
func Width(str string) (width int) {
	for _, l := range strings.Split(str, "\n") {
		w := ansi.PrintableRuneWidth(l)
		if w > width {
			width = w
		}
	}

	return width
}
```

PR #51, opened 2024-03-14
feat(term): ansi: implement wrap and wordwrap

aymanbagabas:
This is based on @muesli's awesome
[reflow](https://github.com/muesli/reflow) library. It uses the ANSI parser state machine to find escape codes and runes. Since it uses the ANSI parser state machine, it supports OSC, DCS, and other sequences.

Related: https://github.com/muesli/reflow/pull/71

Benchmarks running against reflow:

```
› go test -bench=. -benchmem -run=^$ ./ansi/...
goos: darwin
goarch: amd64
pkg: github.com/charmbracelet/x/exp/term/ansi
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkWrap-16                19056892                66.83 ns/op          111 B/op          1 allocs/op
BenchmarkReflowWrap-16           9982068               122.7 ns/op           164 B/op          4 allocs/op
BenchmarkWordwrap-16             7652895               161.3 ns/op           287 B/op          4 allocs/op
BenchmarkReflowWordwrap-16       4818664               254.0 ns/op           287 B/op          4 allocs/op
PASS
```

### DIFF
--- a/exp/term/ansi/wrap.go
+++ b/exp/term/ansi/wrap.go
@@ -5,7 +5,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	. "github.com/charmbracelet/x/exp/term/ansi/parser"
+	"github.com/charmbracelet/x/exp/term/ansi/parser"
 	"github.com/rivo/uniseg"
 )
 
@@ -25,9 +25,8 @@ func Wrap(s string, limit int, preserveSpace bool) string {
 		curWidth     int
 		forceNewline bool
 		gstate       = -1
-		pstate       = GroundState // initial state
+		pstate       = parser.GroundState // initial state
 		b            = []byte(s)
-		i            = 0
 	)
 
 	addNewline := func() {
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

exp/term/ansi/wrap.go
```diff
@@ -0,0 +1,246 @@
...
+						continue
+					}
+				}
+				buf.Write(cluster)
+				curWidth += width
+				gstate = -1 // reset grapheme state otherwise, width calculation might be off
+
+				pstate = GroundState
+				continue
+			} else {
```
mikelorant, line 66: Why not reverse the condition, handle the else first with a continue, then you can do the rest of this outside a conditional. Move the happy path back to the left.

exp/term/ansi/wrap.go
```diff
@@ -0,0 +1,246 @@
...
+					wordLen += width
+					if curWidth+space.Len()+wordLen > limit &&
+						wordLen < limit {
+						addNewline()
+					}
+				}
+
+				pstate = GroundState
+				continue
+			} else {
```
mikelorant, line 190: How about change this around to:
```go
if w := utf8ByteLen(b[i]); w <= 1 {
    // Collect sequence intermediate bytes
    word.WriteByte(b[i])

    break
}

var width int
cluster, _, width, gstate = uniseg.FirstGraphemeCluster(b[i:], gstate)
...
continue
```

This gives you an early exit, then pushes the rest of the left towards the happy path on the left.

exp/term/ansi/wrap.go
```diff
@@ -0,0 +1,246 @@
...
+				// Collect sequence intermediate bytes
+				word.WriteByte(b[i])
+			}
+		case PrintAction, ExecuteAction:
+			r := rune(b[i])
+			if r == '\n' {
```
mikelorant, line 196: Wouldn't this be far more readable as a switch statement instead rather than continually using `else if`?

exp/term/ansi/wrap.go
```diff
@@ -0,0 +1,246 @@
...
+
+	addNewline := func() {
+		buf.WriteByte('\n')
+		curWidth = 0
+		space.Reset()
+	}
+
+	for i < len(b) {
+		state, action := Table.Transition(pstate, b[i])
+		// log.Printf("pstate: %s, state: %s, action: %s, code: %q", StateNames[pstate], StateNames[state], ActionNames[action], b[i])
```
mikelorant, line 159: Temporary? Feel like we need a nice way to set a logging level to enable this rather than comment/uncomment.

exp/term/ansi/wrap.go
```diff
@@ -0,0 +1,246 @@
...
+		cluster  []byte
+		buf      bytes.Buffer
+		word     bytes.Buffer
+		space    bytes.Buffer
+		curWidth int
+		wordLen  int
+		gstate   = -1
+		pstate   = GroundState // initial state
+		b        = []byte(s)
+		i        = 0
```
mikelorant, line 131: This feels too far away from where it is used to be a single letter. Yes, obviously a counter, but the function that uses it is not directly below.

exp/term/ansi/wrap.go
```diff
@@ -0,0 +1,246 @@
+package ansi
+
+import (
+	"bytes"
+	"unicode"
+	"unicode/utf8"
+
+	. "github.com/charmbracelet/x/exp/term/ansi/parser"
```
mikelorant, line 8: Unused, so should this be removed?

mikelorant:
Added some comments, I am fairly intermediate in my Go skills, so take them as observations. I may very well be wrong in my suggestions.

exp/term/ansi/wrap.go
aymanbagabas: I thought about that, however, almost no one is going to log these messages unless they're debugging the flow of `Wrap`. For now, I removed the logging lines 🙂

exp/term/ansi/wrap.go
aymanbagabas: Good point, moved it down to be before the loop

exp/term/ansi/wrap.go
aymanbagabas: It's dot-importing the `parser` package which contains the state machine. Renamed the import and removed the dot-import.

aymanbagabas:
> Added some comments, I am fairly intermediate in my Go skills, so take them as observations. I may very well be wrong in my suggestions.

Thank you @mikelorant for reviewing this one!

mikelorant (approved):

PR #54, opened 2024-03-27
feat: adds getter for program field in TestModel

UncleGedd:
During testing, I need a mechanism to access the `TestModel`'s underlying `program` in order to assign it to a `program` struct that I'm using to send messages back to the TUI during business-logic portions of my app.

In other words, I don't want my test code to have to call `tm.Send`, I'd rather have the existing `Send` calls in my business logic work with the same `program` that the tests are using

Quick and dirty solution is to add a getter for the `program`, but I'm happy to look into other methods.

### DIFF
--- a/exp/teatest/teatest.go
+++ b/exp/teatest/teatest.go
@@ -250,7 +250,7 @@ func (tm *TestModel) Type(s string) {
 	}
 }
 
-// GetProgram gets the TestModel's program
+// GetProgram gets the TestModel's program.
 func (tm *TestModel) GetProgram() *tea.Program {
 	return tm.program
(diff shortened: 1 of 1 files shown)


exp/teatest/teatest.go
```diff
@@ -250,6 +250,11 @@ func (tm *TestModel) Type(s string) {
 	}
 }
 
+// GetProgram gets the TestModel's program
```
caarlos0, line 253:
```suggestion
// GetProgram gets the TestModel's program.
```

caarlos0:
thanks!

PR #57, opened 2024-03-28
feat(term): ansi: move Wrap to Hardwrap and introduce Wrap

aymanbagabas:
ANSI aware text wrapping that breaks word boundaries only when necessary.

Fixes: https://github.com/charmbracelet/lipgloss/pull/275
Fixes: https://github.com/muesli/reflow/issues/43

### DIFF
--- a/exp/term/ansi/wrap.go
+++ b/exp/term/ansi/wrap.go
@@ -9,12 +9,12 @@ import (
 	"github.com/rivo/uniseg"
 )
 
-// Wrap wraps a string or a block of text to a given line length, breaking word
-// boundaries. This will preserve ANSI escape codes and will account for
+// Hardwrap wraps a string or a block of text to a given line length, breaking
+// word boundaries. This will preserve ANSI escape codes and will account for
 // wide-characters in the string.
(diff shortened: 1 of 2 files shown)


exp/term/ansi/wrap.go
```diff
@@ -227,6 +229,153 @@ func Wordwrap(s string, limit int, breakpoints string) string {
 	return buf.String()
 }
 
+// SmartWrap wraps a string or a block of text to a given line length, breaking
+// word boundaries if necessary. This will preserve ANSI escape codes and will
+// account for wide-characters in the string. The breakpoints string is a list
+// of characters that are considered breakpoints for word wrapping. A hyphen
+// (-) is always considered a breakpoint.
+func SmartWrap(s string, limit int, breakpoints string) string {
```
aymanbagabas, line 237: Maybe `SmartWrap` is incorrect since there's a smarter way to wrap the text with less word breakage. Perhaps `Wrap` should be `Hardwrap` and this becomes just `Wrap`.

caarlos0 (approved):

Issue #58, opened 2024-03-31
Wrap adds a new line to strings containing ANSI codes


mikelorant:
When using the `Wrap` function as part of the ANSI package, strings containing ANSI codes that have a printable width equal to the wrap value have a new line appended.

Following code demonstrates this issue:
```go
package main

import (
    "fmt"

    "github.com/charmbracelet/lipgloss"
    "github.com/charmbracelet/x/exp/term/ansi"
)

func main() {
    str := "1"
    red := lipgloss.Color("9")

    style1 := lipgloss.NewStyle().
        Foreground(red).
        Render(str)

    style2 := lipgloss.NewStyle().
        Width(1).
        Render(style1)

    fmt.Println("Lipgloss")
    fmt.Printf("Before Width: '%v'\n", style1)
    fmt.Printf("After Width : '%v'\n", style2)
    fmt.Println()

    fmt.Println("ANSI")
    fmt.Printf("ASCII Width : '%v'\n", ansi.Wrap(str, 1, ""))
    fmt.Printf("ANSI Width  : '%v'\n", ansi.Wrap(style1, 1, ""))
}
```

This is a causing a regression in Lip Gloss when using the style `Width` function to force a string to a specific width.

As the only test function that calls `Wrap` is `TestSmartWrap`, this test case can be added to check for this problem.
```go
{
	name:     "exact",
	input:    "\x1b[91mfoo\x1b[0",
	expected: "\x1b[91mfoo\x1b[0",
	width:    3,
},
```

This test case can be made to pass changing the following line:
https://github.com/charmbracelet/x/blob/62a6965a9fadda46e017a917ec4ded35e97e5c63/exp/term/ansi/wrap.go#L355

To:
```go
} else if curWidth+wordLen > limit {
```

However this causes other tests to fail, so there is some other edge cases to consider.

@aymanbagabas Let me know if there is anything else I need to provide to debug this issue further.

mikelorant:
Figured this one out but have no idea how to fix it.

By the time we get to the 3rd and last printed character, we have moved `i` to a value of 7. However, the final position `i` will get to is `10`.

The logic that matters:
```go
if i < len(b)-1 {
        addNewline()
}
```

The remaining 3 characters are `<ESC>[0`. These 3 characters reset the ANSI sequence. So nothing actually happens, but it is too late, we have already added the new line.

How do we detect there are no more printable characters before we add the new line?

1. Delay this new line **UNTIL** we know another character needs to be printed.
[...]

PR #59, opened 2024-04-01
fix(term): ansi: account for some wrap edge cases

aymanbagabas:
Properly count escape codes, better handling of breakpoints, and only break word/breakpoint when necessary.

Fixes: https://github.com/charmbracelet/x/issues/58

### DIFF
--- a/exp/term/ansi/wrap.go
+++ b/exp/term/ansi/wrap.go
@@ -134,9 +134,11 @@ func Wordwrap(s string, limit int, breakpoints string) string {
 	}
 
 	addWord := func() {
-		if curWidth > 0 && word.Len() > 0 {
-			addSpace()
+		if wordLen == 0 {
+			return
 		}
+
+		addSpace()
 		curWidth += wordLen
 		buf.Write(word.Bytes())
 		word.Reset()
@@ -267,13 +269,11 @@ func Wrap(s string, limit int, breakpoints string) string {
 
 	addWord := func() {
 		addBpoint()
-		if curWidth > 0 && word.Len() > 0 {
-			// We use wordLen to determine if we have a word to add
-			// to the buffer. If wordLen is 0, we don't add spaces at the
-			// beginning of a line.
-			addSpace()
+		if word.Len() == 0 {
+			return
 		}
 
+		addSpace()
 		curWidth += wordLen
 		buf.Write(word.Bytes())
 		word.Reset()
--- a/exp/term/ansi/wrap.go
+++ b/exp/term/ansi/wrap.go
@@ -134,7 +134,7 @@ func Wordwrap(s string, limit int, breakpoints string) string {
 	}
 
 	addWord := func() {
-		if wordLen > 0 {
+		if curWidth > 0 {
 			addSpace()
 		}
 		curWidth += wordLen
--- a/exp/term/ansi/wrap.go
+++ b/exp/term/ansi/wrap.go
@@ -9,6 +9,9 @@ import (
 	"github.com/rivo/uniseg"
 )
 
+// nbsp is a non-breaking space
+const nbsp = 0xA0
+
 // Hardwrap wraps a string or a block of text to a given line length, breaking
 // word boundaries. This will preserve ANSI escape codes and will account for
 // wide-characters in the string.
@@ -106,14 +109,11 @@ func Hardwrap(s string, limit int, preserveSpace bool) string {
 // breakpoints for word wrapping. A hyphen (-) is always considered a
 // breakpoint.
 //
-// Note: breakpoints must be a string of 1-cell wide rune character.
+// Note: breakpoints must be a string of 1-cell wide rune characters.
 func Wordwrap(s string, limit int, breakpoints string) string {
 	if limit < 1 {
 		return s
 	}
-
-	// Add a hyphen to the breakpoints
-	breakpoints += "-"
 
 	var (
 		cluster  []byte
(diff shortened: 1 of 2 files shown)


exp/term/ansi/wrap.go
```diff
@@ -340,23 +339,23 @@ func Wrap(s string, limit int, breakpoints string) string {
 			case runeContainsAny(r, breakpoints):
 				addSpace()
 				addWord()
-				buf.WriteByte(b[i])
-				curWidth++
+				if curWidth+1 <= limit {
```
mikelorant, line 342: Can this be simplified as?
```go
if curWidth < limit {
```

exp/term/ansi/wrap.go
```diff
@@ -340,23 +339,23 @@ func Wrap(s string, limit int, breakpoints string) string {
...
+					bpoint.WriteByte(b[i])
+					break
+				}
+				// If we can't fit the breakpoint in the current line, we treat
+				// it as a word character.
+				fallthrough
 			default:
-				if wordLen+1 > limit {
+				if wordLen >= limit {
+					// If the word is longer than the limit, we break it
```
mikelorant, line 351: This comment doesn't match what they code does.

exp/term/ansi/wrap_test.go
mikelorant, line 1: Great to see you fix the test function names. Was going to point that out earlier. 👍

exp/term/ansi/wrap.go
```diff
@@ -132,10 +134,9 @@ func Wordwrap(s string, limit int, breakpoints string) string {
 	}
 
 	addWord := func() {
-		if word.Len() == 0 {
-			return
+		if wordLen > 0 {
```
aymanbagabas, line 137:
```suggestion
		if curWidth > 0 {
```

exp/term/ansi/wrap.go
```diff
@@ -247,24 +245,35 @@ func Wrap(s string, limit int, breakpoints string) string {
...
 		curWidth += space.Len()
 		buf.Write(space.Bytes())
 		space.Reset()
 	}
 
 	addWord := func() {
-		if word.Len() == 0 {
-			return
+		addBpoint()
+		if wordLen > 0 {
```
aymanbagabas, line 270:
```suggestion
		if curWidth > 0 {
```

exp/term/ansi/wrap_test.go
```diff
@@ -149,19 +149,57 @@ var smartWrapCases = []struct {
 	{
 		name:     "long input2",
 		input:    "Rotated keys for a-good-offensive-cheat-code-incorporated/crypto-line-operating-system.",
-		expected: "Rotated keys for a-good-offensive-cheat-code-incorporated/crypto-line-operat\ning-system.",
```
mikelorant, line 152: This test had me confused on what the expected result was. This was the test case my attempted fixes kept breaking (no pun intended) on.

exp/term/ansi/wrap.go
```diff
@@ -132,10 +134,9 @@ func Wordwrap(s string, limit int, breakpoints string) string {
 	}
 
 	addWord := func() {
-		if word.Len() == 0 {
-			return
+		if curWidth > 0 {
```
aymanbagabas, line 137:
```suggestion
		if curWidth > 0 && word.Len() > 0 {
```

exp/term/ansi/wrap.go
```diff
@@ -247,24 +245,35 @@ func Wrap(s string, limit int, breakpoints string) string {
...
 		curWidth += space.Len()
 		buf.Write(space.Bytes())
 		space.Reset()
 	}
 
 	addWord := func() {
-		if word.Len() == 0 {
-			return
+		addBpoint()
+		if curWidth > 0 {
```
aymanbagabas, line 270:
```suggestion
		if curWidth > 0 && word.Len() > 0 {

```

aymanbagabas:
@mikelorant Pushed a bunch of fixes for edge cases and support for NBSP (non-breaking space) character.

exp/term/ansi/wrap.go
```diff
@@ -336,27 +344,32 @@ func Wrap(s string, limit int, breakpoints string) string {
...
+			case r == '-':
+				fallthrough
 			case runeContainsAny(r, breakpoints):
 				addSpace()
-				addWord()
-				buf.WriteByte(b[i])
-				curWidth++
-			default:
-				if wordLen+1 > limit {
+				if curWidth+wordLen+1 > limit {
```
mikelorant, line 352: Consider dropping the `+1` and making it a `>=` comparison to simplify the logic?

exp/term/ansi/wrap.go
```diff
@@ -199,6 +202,8 @@ func Wordwrap(s string, limit int, breakpoints string) string {
 			case unicode.IsSpace(r):
 				addWord()
 				space.WriteByte(b[i])
+			case r == '-':
```
mikelorant, line 205: Would there be value in having a small helper function `IsDash` to be consistent to how it is done for `IsSpace`?

mikelorant (approved):

mikelorant:
I'm not sure of the commit best practices used by the Charm team, but I generally recommend at this point to the developers I work with to squash their commits (seeing as there are now 8 commits). Your branch, so your rules 😄

aymanbagabas:
> I'm not sure of the commit best practices used by the Charm team, but I generally recommend at this point to the developers I work with to squash their commits (seeing as there are now 8 commits). Your branch, so your rules 😄 

Yes, that was what I'm going for. I don't think the commits should be split since they're all related and address the same issues

exp/term/ansi/wrap.go
aymanbagabas: Perhaps, if we add support to other unicode dash characters 🤔

mikelorant: Something to consider, certainly nothing to stop you merging once you do decide.

mikelorant:
I'd say while there *might* be some minor improvements possible, the current state is worth merging. LGTM.

exp/term/ansi/wrap.go
aymanbagabas, line 352:
```suggestion
				if curWidth+wordLen >= limit {
```

exp/term/ansi/wrap.go
aymanbagabas: Thinking more about this, there are many variants of dashes defined in unicode. People who wish to use custom breakpoints can do so using the `breakpoints` parameter

PR #70, opened 2024-05-06
Simplify term query features

aymanbagabas:
- Remove less used features (reduce maintenance)
- Return query errors
- Customize query timeout


caarlos0 (approved):
good changes, I like it!

PR #71, opened 2024-05-06
XPty: cross-platform Pty support

aymanbagabas:
Uses creack/pty for Unix, and our ConPty package for Windows. Will replace charmbracelet/ssh pty support

- [x] Unix
- [x] Windows


exp/term/xpty/xpty.go
```diff
@@ -0,0 +1,44 @@
...
+	// The command started will have its standard input, output, and error
+	// connected to the PTY.
+	// On Windows, calling Wait won't work since the Go runtime doesn't handle
+	// ConPTY processes correctly. See https://github.com/golang/go/pull/62710.
+	Start(cmd *exec.Cmd) error
+}
+
+var _ io.ReadWriteCloser = XPty(nil)
+
+var _ term.File = XPty(nil)
```
caarlos0, line 44: why not include these 2 interfaces in the XPty interface?

exp/term/xpty/xpty.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/charmbracelet/x/exp/term"
+	"github.com/creack/pty"
+)
+
+// ErrUnsupported is returned when a feature is not supported.
+var ErrUnsupported = pty.ErrUnsupported
+
+// XPty represents a PTY (pseudo-terminal) interface.
+type XPty interface {
```
caarlos0, line 15: could probably call it just Pty, as the pkg name is xpty already 🤔

PR #90, opened 2024-06-03
fix(input): rename keyMod modifier functions

aymanbagabas:


caarlos0 (approved):
nice, `has` indeed makes more sense!

PR #101, opened 2024-06-22
feat: Ansi slice function

sukus21:
Based on the truncate function.
Specify a start point (inclusive) and end point (exclusive) in cell-width units, and get a string back at that exact size.
Handles ansi-codes and multi-length characters.

### DIFF
--- a/ansi/slice.go
+++ b/ansi/slice.go
@@ -1,0 +1,111 @@
+package ansi
+
+import (
+	"bytes"
+
+	"github.com/charmbracelet/x/ansi/parser"
+	"github.com/rivo/uniseg"
+)
+
+// Slice slices a string to a given length, starting from X cell position.
+// This function is aware of ANSI escape codes and will not break them, and
+// accounts for wide-characters (such as East Asians and emojis).
+//
+// If a string is cut in the middle of a wide character, padding (in the
+// form of spaces) is inserted. This is done in order to maintain the width
+// of the input string.
+func Slice(s string, start int, end int) string {
+	if end < start || start == end || s == "" {
+		return ""
+	}
+
+	var cluster []byte
+	var buf bytes.Buffer
+	curPos := 0
+	pstate := parser.GroundState // initial state
+	b := []byte(s)
+
+	// Here we iterate over the bytes of the string and collect printable
+	// characters and runes. We also keep track of the scan position in cells.
+	// Once we reach the given length, we start ignoring characters and only
+	// collect ANSI escape codes until we reach the end of string.
+	for i := 0; i < len(b); i++ {
+		state, action := parser.Table.Transition(pstate, b[i])
+
+		switch action {
+		case parser.PrintAction:
+			// Single/zero width character, fast path
+			if utf8ByteLen(b[i]) <= 1 {
+				if curPos >= start && curPos < end {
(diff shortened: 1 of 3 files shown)


ccoVeille:
I like the code, the quality of comments, the extensive tests. 🥰

ansi/slice.go
```diff
@@ -0,0 +1,111 @@
...
+	curPos := 0
+	pstate := parser.GroundState // initial state
+	b := []byte(s)
+	i := 0
+
+	// Here we iterate over the bytes of the string and collect printable
+	// characters and runes. We also keep track of the scan position in cells.
+	// Once we reach the given length, we start ignoring characters and only
+	// collect ANSI escape codes until we reach the end of string.
+	for i < len(b) {
```
ccoVeille, line 33: The only issue I see with code is the fact, this could lead to loop bomb, if a part of the code is refactored

What about adding an int like `jump` or offset, then use this

```go
i := 0
offset:= 0
for i + offset < len(b) {
       i += offset
       offset=1

// then code set the offset to something larger than 1 when needed, instead of incrementing i

}
```

Another solution might also be to simplify add something like

```go
i := 0
bomb := 0
for i < len(b) {
       bomb++
       if bomb > len(b) {
            break
       }


// keep current code in the PR unchanged

}
```

sukus21: I am not sure I understand what you mean by a "loop bomb", I have never encountered that term before. And wouldn't the unit tests catch any potential issues a refactor might bring along with it?

ccoVeille: An infinite loop if you prefer.

Yes, test could catch it but it can take time to understand the issue in test if they have a timeout, and could eat CPU resources on production or CI code if it gets unnoticed

Let me try to spot how it could happen

Look at like 44 or 92 the `i++`. if someone remove the line, the for loop will never end.

Another use case could be someone who adds an if just before the line 44, use a continue in it, but forget the `i++`

ccoVeille: It's a detail, but when I see a for without a clear way to exit the loop, I'm worried

sukus21: I understand. I will try and think of a solution

ccoVeille: Any solution that prevent an infinite loop is good to me. If I had provided you an idea about the offset variable it was for limiting the refactoring

ccoVeille (approved):
I find the code clearer now, BTW.

Thanks for baring with my remarks

aymanbagabas (approved):
LGTM, thanks @sukus21 for your contribution!

aymanbagabas:
@sukus21, I've reverted this PR. I think we need to think through the implication of adding space paddings to the sliced string. I'm not sure if that's the right approach here. Take `string` slicing for example, something like `s[1:3]` will slice the string as if it was a byte slice and breaking utf8 codepoints.
Using something like [`golang.org/x/exp/utf8string`](https://pkg.go.dev/golang.org/x/exp/utf8string) `Slice` accounts for runes and returns the sliced string with respect to utf8 codepoints (runes) and the `start` and `end` arguments resemble utf8 rune indices.

With that being said, I'm not sure how we _should_ handle slicing wide characters and I don't think padding with spaces is the right solution, especially if `end > StringWidth(s)`.

PR #105, opened 2024-06-28
Draft: added ansi slice function

sukus21:
Re-apply features from #101, reopened per request from @aymanbagabas.

**Original description:**
Based on the truncate function.
Specify a start point (inclusive) and end point (exclusive) in cell-width units, and get a string back at that exact size.
Handles ansi-codes and multi-length characters.

### DIFF
--- a/ansi/slice.go
+++ b/ansi/slice.go
@@ -1,0 +1,111 @@
+package ansi
+
+import (
+	"bytes"
+
+	"github.com/charmbracelet/x/ansi/parser"
+	"github.com/rivo/uniseg"
+)
+
+// Slice slices a string to a given length, starting from X cell position.
+// This function is aware of ANSI escape codes and will not break them, and
+// accounts for wide-characters (such as East Asians and emojis).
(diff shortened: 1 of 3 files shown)


sukus21:
The reason I made this code in the first place, was in an attempt to recreate the example shown for Bubble Tea on the [Charmbracelet website](https://charm.sh). While truncate can cut off the end of a string, I saw no code that could cut the beginning off a string. So I decided to make the `Slice` function.

As for the decision to pad with whitespace when splitting on wide characters, that was for my own usecase. I did not see the point of the user manually having to check if the character that was split on was wide, and them then having to insert padding themselves. That would mean essentially recreating the functionality of the `Slice` function, after it has already been called.
I am not trying to argue that this is the best solution, but it was the one that fit the use case I had in mind when writing it. 

An alternative could be to let the user choose how wide characters are handled:
* Include the wide character
* Exclude the wide character
* Pad the string with whitespace, to match expected length

PR #115, opened 2024-07-08
fix(ansi): truncate: truncate string when input length is equal to output

aymanbagabas:
```
goos: darwin
goarch: arm64
pkg: github.com/charmbracelet/x/ansi
                  │   old.txt   │              new.txt               │
                  │   sec/op    │   sec/op     vs base               │
TruncateString-16   24.96n ± 1%   25.83n ± 2%  +3.49% (p=0.001 n=10)

                  │  old.txt   │            new.txt             │
                  │    B/op    │    B/op     vs base            │
TruncateString-16   65.00 ± 0%   65.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                  │  old.txt   │            new.txt             │
                  │ allocs/op  │ allocs/op   vs base            │
TruncateString-16   1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

Fixes: https://github.com/charmbracelet/lipgloss/issues/324
Fixes: https://github.com/charmbracelet/x/pull/106

CC/ @mikelorant @jbcpollak

### DIFF
--- a/ansi/truncate.go
+++ b/ansi/truncate.go
@@ -12,6 +12,10 @@ import (
 // This function is aware of ANSI escape codes and will not break them, and
 // accounts for wide-characters (such as East Asians and emojis).
 func Truncate(s string, length int, tail string) string {
+	if sw := StringWidth(s); sw <= length {
+		return s
+	}
+
 	tw := StringWidth(tail)
 	length -= tw
 	if length < 0 {
--- a/ansi/truncate_test.go
+++ b/ansi/truncate_test.go
@@ -16,6 +16,7 @@ var tcases = []struct {
 	{"equalascii", "one", ".", 3, "one"},
 	{"equalemoji", "on👋", ".", 3, "on."},
 	{"equalcontrolemoji", "one\x1b[0m", ".", 3, "one\x1b[0m"},
+	{"truncate_tail_greater", "foo", "...", 5, "foo"},
 	{"simple", "foobar", "", 3, "foo"},
 	{"passthrough", "foobar", "", 10, "foobar"},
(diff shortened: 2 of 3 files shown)


ansi/truncate.go
```diff
@@ -12,8 +12,13 @@ import (
 // This function is aware of ANSI escape codes and will not break them, and
 // accounts for wide-characters (such as East Asians and emojis).
 func Truncate(s string, length int, tail string) string {
-	tw := StringWidth(tail)
-	length -= tw
+	if sw := StringWidth(s); sw == length {
```
ccoVeille, line 15: Unsure to understand why you need to check the exact length.

Then you go through all the code.

I would have expected something like this

```diff
-if sw := StringWidth(s); sw == length
+if sw := StringWidth(s); sw <= length {
+       return s
+}
-        tail = "" 
-} else { 
-          tw := StringWidth(tail) 
-          length -= tw 
-}
+tw := StringWidth(tail) 
+length -= tw 
```

aymanbagabas: Ah, good catch! I'll update the commit and benchmarks

ccoVeille: You will have yo update the commit message and add unit test for this.


something like

"foo", 5, "..."

aymanbagabas:
k, was able to get nearly the same performance by improving the implementation of `StringWidth`

ansi/truncate.go
aymanbagabas: added test case

ccoVeille (approved):

caarlos0 (approved):

PR #132, opened 2024-08-03
Introduce ansi.DecodeSequence

aymanbagabas:
This new `ansi.DecodeSequence` replaces the existing ANSI parser by providing a simpler, static method of decoding a single sequence. Instead of passing a `ansi.ParserDispatcher` func, use `ansi.DecodeSequence` to segment the given input into sequences without allocating any memory. This also adds support to sequence widths by using the uniseg library to decode graphemes and report sequence widths instead of manually using uniseg in the dispatcher func in the parser.

See [parserlog2/main.go](https://github.com/charmbracelet/x/blob/ansi/decode-seq/examples/parserlog2/main.go) for an example of how this works compared to [parserlog/main.go](https://github.com/charmbracelet/x/blob/ansi/decode-seq/examples/parserlog/main.go)

some benchmarks:
```
goos: darwin
goarch: arm64
pkg: github.com/charmbracelet/x/ansi
BenchmarkDecodeSequence-16       8324944               138.2 ns/op             0 B/op          0 allocs/op
BenchmarkDecodeParser-16         6610164               181.8 ns/op            96 B/op          2 allocs/op
PASS
ok      github.com/charmbracelet/x/ansi 2.857s
?       github.com/charmbracelet/x/ansi/parser  [no test files]
```

### DIFF
--- a/ansi/parser/transition_table.go
+++ b/ansi/parser/transition_table.go
@@ -82,6 +82,7 @@ func r(start, end byte) []byte {
 //     instead use it to denote sub-parameters.
 //   - Support dispatching SosPmApc sequences.
 //   - The DEL (0x7F) character is executed in the Ground state.
+//   - The DEL (0x7F) character is collected in the DcsPassthrough string state.
 //   - The ST C1 control character (0x9C) is executed and not ignored.
 func GenerateTransitionTable() TransitionTable {
 	table := NewTransitionTable(DefaultTableSize)
@@ -212,7 +213,7 @@ func GenerateTransitionTable() TransitionTable {
 	table.AddOne(0x19, DcsStringState, PutAction, DcsStringState)
 	table.AddRange(0x1C, 0x1F, DcsStringState, PutAction, DcsStringState)
 	table.AddRange(0x20, 0x7E, DcsStringState, PutAction, DcsStringState)
-	table.AddOne(0x7F, DcsStringState, IgnoreAction, DcsStringState)
+	table.AddOne(0x7F, DcsStringState, PutAction, DcsStringState)
 	table.AddRange(0x80, 0xFF, DcsStringState, PutAction, DcsStringState) // Allow Utf8 characters by extending the printable range from 0x7F to 0xFF
 	// ST, CAN, SUB, and ESC terminate the sequence
 	table.AddOne(0x1B, DcsStringState, DispatchAction, EscapeState)
--- a/ansi/parser/transition_table.go
+++ b/ansi/parser/transition_table.go
@@ -81,6 +81,7 @@ func r(start, end byte) []byte {
 //   - We don't ignore 0x3A (':') when building Csi and Dcs parameters and
 //     instead use it to denote sub-parameters.
 //   - Support dispatching SosPmApc sequences.
+//   - The DEL (0x7F) character is executed in the Ground state.
 func GenerateTransitionTable() TransitionTable {
 	table := NewTransitionTable(DefaultTableSize)
 	table.SetDefault(NoneAction, GroundState)
@@ -116,7 +117,8 @@ func GenerateTransitionTable() TransitionTable {
 	table.AddRange(0x00, 0x17, GroundState, ExecuteAction, GroundState)
 	table.AddOne(0x19, GroundState, ExecuteAction, GroundState)
 	table.AddRange(0x1C, 0x1F, GroundState, ExecuteAction, GroundState)
-	table.AddRange(0x20, 0x7F, GroundState, PrintAction, GroundState)
+	table.AddRange(0x20, 0x7E, GroundState, PrintAction, GroundState)
+	table.AddOne(0x7F, GroundState, ExecuteAction, GroundState)
 
 	// EscapeIntermediate
 	table.AddRange(0x00, 0x17, EscapeIntermediateState, ExecuteAction, EscapeIntermediateState)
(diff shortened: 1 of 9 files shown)


ansi/parser/transition_table.go
```diff
@@ -116,7 +117,8 @@ func GenerateTransitionTable() TransitionTable {
 	table.AddRange(0x00, 0x17, GroundState, ExecuteAction, GroundState)
 	table.AddOne(0x19, GroundState, ExecuteAction, GroundState)
 	table.AddRange(0x1C, 0x1F, GroundState, ExecuteAction, GroundState)
-	table.AddRange(0x20, 0x7F, GroundState, PrintAction, GroundState)
+	table.AddRange(0x20, 0x7E, GroundState, PrintAction, GroundState)
+	table.AddOne(0x7F, GroundState, ExecuteAction, GroundState)
```
ccoVeille, line 122: I would add constant for each key


```suggestion
	table.AddOne(keyDelete, GroundState, ExecuteAction, GroundState)
```

```go
const keyDelete = 0x7F
```

It would avoid these magical numbers

ansi/parser_decode_test.go
```diff
@@ -0,0 +1,337 @@
...
+		},
+		{
+			name:  "ascii space",
+			input: []byte(" "),
+			expected: []expectedSequence{
+				{seq: []byte{' '}, n: 1, width: 1},
+			},
+		},
+		{
+			name:  "ascii del",
```
ccoVeille, line 45:
```suggestion
			name:  "ASCII delete key",
```

ansi/parser_decode_test.go
```diff
@@ -0,0 +1,337 @@
...
+			input: []byte{DEL},
+			expected: []expectedSequence{
+				{seq: []byte{DEL}, n: 1},
+			},
+		},
+		{
+			name:  "del in the middle of utf8 string",
```
ccoVeille, line 52:
```suggestion
			name:  "DEL in the middle of UTF-8 string", 
```

ansi/parser_decode_test.go
```diff
@@ -0,0 +1,337 @@
...
+			name:  "st in the middle of dcs",
+			input: []byte("\x1bP1;2+xa\x9cb\x1b\\"),
+			expected: []expectedSequence{
+				{seq: []byte("\x1bP1;2+xa\x9c"), n: 9},
+				{seq: []byte{'b'}, n: 1, width: 1},
+				{seq: []byte("\x1b\\"), n: 2},
+			},
+		},
+		{
+			name:     "csi",
```
ccoVeille, line 77:
```suggestion
			name:     "CSI",
```

But I have no idea, what  is a CSI

ansi/parser_decode_test.go
```diff
@@ -0,0 +1,337 @@
...
+		{
+			name:  "double st 8-bit",
+			input: []byte("\x9c\x9c"),
+			expected: []expectedSequence{
+				{seq: []byte{ST}, n: 1},
+				{seq: []byte{ST}, n: 1},
+			},
+		},
+		{
+			name:  "ascii printables",
```
ccoVeille, line 150:
```suggestion
			name:  "ASCII printables",
```

ansi/parser_decode_test.go
```diff
@@ -0,0 +1,337 @@
...
+		},
+		{
+			name:  "rune",
+			input: []byte("👋"),
+			expected: []expectedSequence{
+				{seq: []byte("👋"), n: 4, width: 2},
+			},
+		},
+		{
+			name:  "multiple sequences with utf8 and double esc",
```
ccoVeille, line 176:
```suggestion
			name:  "multiple sequences with UTF-8 and double ESC",
```

ansi/parser_decode.go
```diff
@@ -0,0 +1,320 @@
...
+//		seq, width, n, newState := DecodeSequence(input, state, p)
+//		log.Printf("seq: %q, width: %d", seq, width)
+//		state = newState
+//		input = input[n:]
+//	}
+func DecodeSequence[T string | []byte](b T, state byte, p *Parser) (seq T, width int, n int, newState byte) {
+	// log.Printf("DecodeSequence(%q, %d)", b, state)
+	// defer func() {
+	// 	log.Printf("DecodeSequence(%q, %d) -> (%q, %d, %d, %d)", b, state, seq, width, n, newState)
+	// }()
```
ccoVeille, line 39: Left behind debug ?

ansi/parser_decode_test.go
```diff
@@ -0,0 +1,337 @@
...
+			name:  "del in the middle of utf8 string",
+			input: []byte{'a', DEL, 'b'},
+			expected: []expectedSequence{
+				{seq: []byte{'a'}, n: 1, width: 1},
+				{seq: []byte{DEL}, n: 1},
+				{seq: []byte{'b'}, n: 1, width: 1},
+			},
+		},
+		{
+			name:  "del in the middle of dcs",
```
ccoVeille, line 61:
```suggestion
			name:  "DEL in the middle of DCS",
```

ansi/parser/transition_table.go
aymanbagabas: Wanted to import `ansi` and use the consts from there, but that would introduce an import cycle, this is fine since all of these are single bytes i.e. have a range from 0x00-0xFF.

PR #144, opened 2024-08-09
feat(editor): open at the end of the line

caarlos0:
allows to tell the editor to open the file in the end of the line, should work with and without specifying which line.

### DIFF
--- a/editor/editor.go
+++ b/editor/editor.go
@@ -17,7 +17,14 @@ const defaultEditor = "nano"
 type Option func(editor, filename string) (args []string, pathInArgs bool)
 
 // OpenAtLine opens the file at the given line number in supported editors.
+//
+// Deprecated: use Line instead.
 func OpenAtLine(number uint) Option {
+	return Line(number)
+}
+
+// Line opens the file at the given line number in supported editors.
+func Line(number uint) Option {
 	plusLineEditors := []string{"vi", "vim", "nvim", "nano", "emacs", "kak", "gedit"}
 	return func(editor, filename string) ([]string, bool) {
 		for _, e := range plusLineEditors {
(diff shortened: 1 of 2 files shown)


editor/editor.go
```diff
@@ -35,6 +35,15 @@ func OpenAtLine(number uint) Option {
 	}
 }
 
+// OpenAtEndOfLine opens the file at the end of the line.
+func OpenAtEndOfLine(editor, _ string) (args []string, pathInArgs bool) {
```
aymanbagabas, line 39: Shouldn't this return an `Option`?

caarlos0: It **is** an option, since there's no arguments, you can simply pass the function

aymanbagabas: oh my bad, i overlooked that part 😅

aymanbagabas: perhaps we could make it explicitly return an `Option`

caarlos0: I mean... there's no reason for it as unlike open at line, it doesn't take any options...

there's no great reason against it either, I think

editor/editor.go
```diff
@@ -17,7 +17,14 @@ const defaultEditor = "nano"
...
 
 // OpenAtLine opens the file at the given line number in supported editors.
+//
+// Deprecated: use Line instead.
 func OpenAtLine(number uint) Option {
+	return Line(number)
+}
+
+// Line opens the file at the given line number in supported editors.
+func Line(number uint) Option {
```
aymanbagabas, line 27: Merely a suggestion to make this option clearer 🤔

```suggestion
// LineNr is an option that opens the file at the given line number in supported editors.
func LineNr(number uint) Option {
```

aymanbagabas (approved):

PR #160, opened 2024-08-20
feat(ansi): add cursor and mouse pointer shape sequences

aymanbagabas:
This adds `CSI Ps SP q` (DECSCUSR) and `OSC 22` sequences to change the terminal cursor shape, and the mouse pointer in shape.

### DIFF
--- a/ansi/cursor.go
+++ b/ansi/cursor.go
@@ -188,3 +188,48 @@ const SaveCursorPosition = "\x1b[s"
 //
 // See: https://vt100.net/docs/vt510-rm/SCORC.html
 const RestoreCursorPosition = "\x1b[u"
+
+// SetCursorStyle (DECSCUSR) returns a sequence for changing the cursor style.
+//
+//	CSI Ps SP q
+//
+// Where Ps is the cursor style:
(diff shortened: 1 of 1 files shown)


ansi/cursor.go
```diff
@@ -188,3 +188,48 @@ const SaveCursorPosition = "\x1b[s"
...
 // See: https://vt100.net/docs/vt510-rm/SCORC.html
 const RestoreCursorPosition = "\x1b[u"
+
+// SetCursorStyle (DECSCUSR) returns a sequence for changing the cursor style.
+//
+//	CSI Ps SP q
+//
+// Where Ps is the cursor style:
+//
+//	0: Blinking block
```
caarlos0, line 198: Both 0 and 2 are blinking blocks?

caarlos0 (approved):

ansi/cursor.go
aymanbagabas: Yes, both 0 and 1 are blinking blocks. See https://vt100.net/docs/vt510-rm/DECSCUSR.html

PR #197, opened 2024-09-17
feat: add teatest/v2 to use bubbletea@v2-exp

aymanbagabas:
Will keep this open until we push a bubbletea/v2 release. I think it'd be good to keep both versions available and in separate packages.

Supersedes: https://github.com/charmbracelet/x/pull/191


caarlos0:
lgtm, needs to run the scripts to update the CI/dependabot though

caarlos0 (approved):

PR #203, opened 2024-10-02
A quick fix to #202

### LINKED ISSUE #202: Logic error from a typo in `open_unix.go`
I'm pretty sure the `==` on line 22 should be `!=`. I'm guessing these lines don't get executed often if you haven't noticed. :)

https://github.com/charmbracelet/x/blob/a4978c8e603ab127212de0270d1109a6f3be2222/exp/open/open_unix.go#L21-L26

I'm also not sure that it makes sense to keep that `if` block inside the conditional on whether `xdg-open` exists, but that's a separate issue.

ilyagr:
The important part is the change of `==` to `!=`. I also moved the if block to a more sensible place.

I have not tested this.

Fixes #202


caarlos0:
ah, good catch! thanks!

PR #217, opened 2024-10-17
BREAKING: Method cell calculation and refactor modes and cursor

aymanbagabas:
- Implement ANSI mode types
- Fix the order of `ansi.MoveCursor`
- Implement `ansi.Method` to respect the cell widths on various `ansi` operations

```go
ansi.StringWidth("👩‍👩‍👧‍👧") // 2
ansi.GraphemeWidth.StringWidth("👩‍👩‍👧‍👧") // 2
ansi.WcWidth.StringWidth("👩‍👩‍👧‍👧") // 6
```


ansi/parser_decode.go
```diff
@@ -67,7 +68,113 @@ const (
...
+// the packed command value that contains intermediate and marker characters.
+// In the case of a OSC sequence, the cmd will be the OSC command number. Use
+// [Cmd] and [Param] types to unpack command intermediates and markers as well
+// as parameters.
+//
+// Zero [p.Cmd] means the CSI, DCS, or ESC sequence is invalid. Moreover, checking the
+// validity of other data sequences, OSC, DCS, etc, will require checking for
+// the returned sequence terminator bytes such as ST (ESC \\) and BEL).
+//
+// We store the command byte in [p.Cmd] in the most significant byte, the
```
bashbunni, line 154: Heads up this isn't linking properly in the godoc 
image

Not sure if this should be `[Parser]` or `[Cmd]` instead

ansi/cursor.go
```diff
@@ -153,7 +153,7 @@ func CursorPreviousLine(n int) string {
 //	CSI n ; m H
 //
 // See: https://vt100.net/docs/vt510-rm/CUP.html
-func MoveCursor(row, col int) string {
+func MoveCursor(col, row int) string {
```
bashbunni, line 156: why change the order?

aymanbagabas: It's a common pattern to use the `x, y int` or `col, row int` than the opposite, it was confusing and led me to swap the values sometimes. It was implemented this way because the [ANSI sequence (CUP)](https://vt100.net/docs/vt510-rm/CUP.html) takes a row then a column in that order.

ansi/parser_decode.go
aymanbagabas: Good catch, this supposed to be `[Cmd]`

ansi/cursor.go
bashbunni: Ah ok that totally makes sense. Thanks for explaining!

bashbunni: Ah ok that totally makes sense. Thanks for explaining!

PR #225, opened 2024-10-28
fix: compat layer for ansi 0.3->0.4 migration

caarlos0:
currently, updating to v0.4.0 is a breaking change... so it'll break any bubbletea app on latest bubbletea version.

I think we can roll this for a while and then remove these deprecated options... wdyt?

### DIFF
--- a/ansi/cursor.go
+++ b/ansi/cursor.go
@@ -163,7 +163,7 @@ func SetCursorPosition(col, row int) string {
 	return "\x1b[" + strconv.Itoa(row) + ";" + strconv.Itoa(col) + "H"
 }
 
-// SetCursorPosition (CUP) returns a sequence for setting the cursor to the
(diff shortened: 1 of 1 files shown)


meowgorithm (approved):

aymanbagabas (approved):

ansi/cursor.go
```diff
@@ -163,10 +163,28 @@ func SetCursorPosition(col, row int) string {
 	return "\x1b[" + strconv.Itoa(row) + ";" + strconv.Itoa(col) + "H"
 }
 
+// SetCursorPosition (CUP) returns a sequence for setting the cursor to the
```
aymanbagabas, line 166:
```suggestion
// MoveCursor (CUP) returns a sequence for setting the cursor to the
```

PR #229, opened 2024-10-30
fix: set cursor style sequence

caarlos0:
extra space :)


aymanbagabas (approved):
Good catch!

aymanbagabas:
Whoops, we should revert this

aymanbagabas:
> ohhhh right... so the parser is not dealing with this properly then?

The handler was invalid, opened a PR

caarlos0:
i realized it was missing the ' ' intermediate shift instantly after i commented :kek:

PR #230, opened 2024-10-30
feat: add OSC 9 notifications

caarlos0:
not sure if we want to add this, as most terms don't support it... wdyt?

### DIFF
--- a/ansi/notification.go
+++ b/ansi/notification.go
@@ -8,5 +8,5 @@ package ansi
 //
 // See: https://iterm2.com/documentation-escape-codes.html
 func Notify(s string) string {
-	return "\x1b]52;" + s + "\x07"
+	return "\x1b]9;" + s + "\x07"
 }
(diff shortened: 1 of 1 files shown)


ansi/notification.go
```diff
@@ -0,0 +1,12 @@
+package ansi
+
+// Notify sends a desktop notification using iTerm's OSC 9.
+//
+//	OSC 9 ; Mc ST
```
aymanbagabas, line 5:
```suggestion
//	 OSC 9 ; Mc ST
//	 OSC 9 ; Mc BEL
```

ansi/notification.go
```diff
@@ -0,0 +1,12 @@
...
+//
+// Where Mc is the notification body.
+//
+// See: https://iterm2.com/documentation-escape-codes.html
+func Notify(s string) string {
+	return "\x1b]52;" + s + "\x07"
```
aymanbagabas, line 11:
```suggestion
	return "\x1b]9;" + s + "\x07"
```

aymanbagabas (approved):

ansi/notification.go
caarlos0: lol forgot to change only the **most** important part

Issue #248, opened 2024-11-07
input: Driver.Cancel Doesn't Work On Windows


CannibalVox:
**Describe the bug**

The Input package has a Driver type that has a ReadEvents method and a Cancel method that seems like it is supposed to cause in-flight ReadEvents calls to die and return an error.  This does not work on Windows, because the underlying engine of ReadEvents is the `github.com/erikgeiser/coninput ReadConsoleInput` method, whose underlying engine is Windows' `ReadConsoleInputW` syscall.  This syscall has no cancellation or timeout mechanism and will hang forever until it receives some sort of input.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS Windows 10
 - Shell Windows CMD
 - Terminal Emulator Windows Terminal

**To Reproduce**
1. Attempt to retrieve background color with lipgloss v2

**Source Code**
lipgloss v2's standalone color example

**Expected behavior**
Either retrieving the background successfully or time out

CannibalVox:
After digging around a bit:

* The windows Driver seems to be built to specifically handle key and mouse input only, with the idea that full blocking on a goroutine waiting for that stuff makes sense
* Unfortunately, non-windows Driver handles EVERYTHING.  Calling the Driver with an ANSI query will inevitably work on non-windows and fail on windows
* The windows CancelReader seems to handle everything appropriately (? not sure on this) but nothing actually calls the windows CancelReader. The stdin handle is used by ReadConsoleInput, but the methods are not ever actually called.  This seems like an issue.

CannibalVox:
I have verified locally that removing the windows-specific code for Driver.ReadEvents fixes this issue (and also successfully retrieves the background color in the preview version of windows terminal) but I don't know what knock-on consequences that might have elsewhere. I see that bubbletea v2 has its own version of this code and has a `readInputs` method that calls ReadEvents and I suspect that that code actually intends to call the blocking ReadConsoleInput method, so I imagine that both are necessary.

aymanbagabas:
Hi @CannibalVox, thank you for reaching out and reporting this issue. Indeed, Bubble Tea has its own version of `x/input`. We're planning on deprecating `x/input` since the code moved to Bubble Tea v2.

[...]

PR #249, opened 2024-11-08
add ReadQueryResponses method that will read ansi responses on windows

CannibalVox:
See #248 

Non-windows systems have one way to read events from the terminal: Driver.ReadEvents calls Driver.readEvents, which pulls data from the CancellableReader and parses ANSI data into events.

Windows systems additionally have code that calls into the Windows ReadConsoleInput syscall, which can only read mouse and keyboard events, and will block until an event is received. This syscall cannot be cancelled or timed out.

Currently, there is no way to call the former (which works great!) on Windows.  Bubbletea v2 seems to use code that does the latter from an input pump, and I assume it works great in that context.  However, ReadEvents is used both for the bubbletea input pump and reading ANSI query responses in lipgloss v2.  When used for the latter, the Windows code will hang because ReadConsoleInput isn't used for pulling ANSI query responses and can't be cancelled.

As a result, I would like to split the Driver API into two methods: ReadEvents, to be used from input pumps, and ReadQueryResponses, to be used from queryTerminal & equivalents.  

I've verified that when plugged in to use the new method from windows terminal & standalone CMD terminals, lipgloss v2's `BackgroundColor` will successfully parse the primary DA response and immediately return a null background color.  When used from the windows terminal preview version, lipgloss v2's `BackgroundColor` will actually return the correct background color.

### DIFF
--- a/input/driver.go
+++ b/input/driver.go
@@ -53,6 +53,13 @@ func NewDriver(r io.Reader, term string, flags int) (*Driver, error) {
 	return d, nil
 }
 
+// ReadQueryResponses is equivalent to ReadEvents on non-Windows systems.  However,
+// on windows systems this method will pull ansi responses from the tty instead of trying
+// to pull key/mouse input from windows' ReadConsoleInput syscall, which is what ReadEvents does
+func (d *Driver) ReadQueryResponses() (e []Event, err error) {
+	return d.readEvents()
+}
+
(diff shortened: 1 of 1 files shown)


CannibalVox:
@aymanbagabas put together the actually correct fix for this in https://github.com/charmbracelet/lipgloss/pull/429

PR #250, opened 2024-11-08
feat(teatest/v2): use virtual terminal

caarlos0:
This changes x/exp/teatest/v2 to use x/vt as its input/output.

This is effectively a breaking change, which should be ok as teatest is still experimental.

- `.Output()` will now return a text representation of the current output, instead of what was written since the last call to it
- `.FinalOutput()` is pretty waits for the program to exit and then calls `.Output()`

IMHO this is good, let me know what you think!


aymanbagabas:
@caarlos0 Could you split the changes into 2 PRs? It's hard to look at when for are in the same one 😄

aymanbagabas:
Could you rebase this onto `vt`

exp/teatest/v2/teatest.go
```diff
@@ -118,18 +122,28 @@ type TestModel struct {
 
 // NewTestModel makes a new TestModel which can be used for tests.
 func NewTestModel(tb testing.TB, m tea.Model, options ...TestOption) *TestModel {
+	var opts TestModelOptions
+	for _, opt := range options {
+		opt(&opts)
+	}
+	if opts.size.Width == 0 {
+		opts.size.Width, opts.size.Height = 70, 40
```
aymanbagabas, line 130:
```suggestion
		opts.size.Width, opts.size.Height = 80, 24
```

PR #263, opened 2024-11-13
fix(vt): lock terminal.Read/Write/String

caarlos0:
- sync.Mutex to lock Read, Write, and String terminal methods


aymanbagabas:
This is already in the `vt` branch

caarlos0:
ah cool, thank you! 🙏🏻

PR #269, opened 2024-11-15
feat: Add sixel sequence object

CannibalVox:
In https://github.com/charmbracelet/lipgloss/pull/436 @aymanbagabas requested that I move ANSI escape code logic to this repository. The logic for building the palette and pixels seem complex enough that they should remain in lipgloss(?) but the basic writing of ANSI envelope is now here in x/ansi.

### DIFF
--- a/cellbuf/go.mod
+++ b/cellbuf/go.mod
@@ -3,7 +3,7 @@ module github.com/charmbracelet/x/cellbuf
 go 1.18
 
 require (
-	github.com/charmbracelet/colorprofile v0.1.10
+	github.com/charmbracelet/colorprofile v0.2.0
 	github.com/charmbracelet/x/ansi v0.8.0
 	github.com/charmbracelet/x/term v0.2.1
 	github.com/mattn/go-runewidth v0.0.16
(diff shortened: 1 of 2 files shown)


ansi/sixel.go
```diff
@@ -0,0 +1,73 @@
+package ansi
+
+import (
+	"bytes"
+	"strconv"
+)
+
+type SixelSequence struct {
```
aymanbagabas, line 8: Sixel is just a DCS command. This should be a `type SixelImage struct` that conforms to `image.Image`.

aymanbagabas (commented):
It would be great to have our own Sixel implementation under `ansi/sixel`, however. That seems like a longer initiative. For now, we could get by using https://github.com/mattn/go-sixel

ansi/sixel.go
CannibalVox: Sixels don't really lend themselves to random access by pixel- Image.At() would basically have to be O(N) where N is the number of bytes in the encoded sixel image.

caarlos0 (approved):

PR #324, opened 2025-01-13
feat(ansi): iterm2: add iTerm2 Inline Image Protocol

aymanbagabas:
This implements the iTerm2 Inline Image Protocol. The protocol allows applications to display images inline in the terminal. The protocol is documented at https://iterm2.com/documentation-images.html.

### DIFF
--- a/ansi/iterm2/file.go
+++ b/ansi/iterm2/file.go
@@ -1,0 +1,130 @@
+package iterm2
+
+import (
+	"strconv"
+	"strings"
+)
+
+// Auto is a constant that represents the "auto" value.
+const Auto = "auto"
+
+// Cells returns a string that represents the number of cells. This is simply a
+// wrapper around strconv.Itoa.
+func Cells(n int) string {
+	return strconv.Itoa(n)
+}
+
+// Pixels returns a string that represents the number of pixels.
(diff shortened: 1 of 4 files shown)


ansi/iterm2/file.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Percent returns a string that represents the percentage.
+func Percent(n int) string {
+	return strconv.Itoa(n) + "%"
+}
+
+// file represents the optional arguments for the iTerm2 Inline Image Protocol.
+//
+// See https://iterm2.com/documentation-images.html
+type file struct {
```
caarlos0, line 30: maybe call this image instead?

caarlos0 (approved):
looks good, maybe worth making the names more specific, as they are images specifically, not general files.

ansi/iterm2/file.go
aymanbagabas: The iTerm2 protocol is originally a file transfer protocol. If the `inline` option is disabled or omitted, the file content is then downloaded to the user's Downloads folder

caarlos0: oh, interesting 🤔 

thanks for clarifying!

PR #327, opened 2025-01-14
feat(ansi): implement DecodeColor function

aymanbagabas:
This implements reading SGR colors according to the ITU T.416 and ECMA standards. It supports reading colors in the following formats: RGB, CMY, CMYK, and indexed colors. The function also supports reading color values separated by semicolons (;) and colons (:).

### DIFF
--- a/ansi/color.go
+++ b/ansi/color.go
@@ -216,8 +216,8 @@ func toRGBA(r, g, b uint32) (uint32, uint32, uint32, uint32) {
 // returned color and will be ignored.
 //
 // This implementation includes a few modifications to the specs:
-//  1. Support for legacy color values separated by semicolons (;) with respect to RGB, CMY, CMYK, and indexed colors
-//  2. Support ignoring and omitting the color space id (second parameter)
+//  1. Support for legacy color values separated by semicolons (;) with respect to RGB, and indexed colors
+//  2. Support ignoring and omitting the color space id (second parameter) with respect to RGB colors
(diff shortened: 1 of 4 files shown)


ansi/color.go
```diff
@@ -194,3 +194,173 @@ func toRGBA(r, g, b uint32) (uint32, uint32, uint32, uint32) {
...
+//
+// The parameters can be separated by semicolons (;) or colons (:). Mixing
+// separators is not allowed.
+//
+// The specs supports defining a color space id, a color tolerance value, and a
+// tolerance color space id. However, these values have no effect on the
+// returned color and will be ignored.
+//
+// This implementation includes a few modifications to the specs:
+//  1. Support for legacy color values separated by semicolons (;) with respect to RGB, CMY, CMYK, and indexed colors
```
aymanbagabas, line 219:
```suggestion
//  1. Support for legacy color values separated by semicolons (;) with respect to RGB, and indexed colors
```

ansi/color.go
```diff
@@ -194,3 +194,173 @@ func toRGBA(r, g, b uint32) (uint32, uint32, uint32, uint32) {
...
+//  2. Support ignoring and omitting the color space id (second parameter)
```
aymanbagabas, line 220:
```suggestion
//  2. Support ignoring and omitting the color space id (second parameter) with respect to RGB colors
```

caarlos0 (approved):
nice, was also thinking of moving this out of sequin.

also this impl looks better than the one there I think!

PR #301, opened 2024-12-10
feat(ansi): TruncateLeft

caarlos0:
Truncate returns the first N chars of the string.
TruncateLeft skips the first N chars of the string, returning the rest.

pretty sure it can be improved.

will be useful for implementing an horizontal scroll in viewport, e.g. https://github.com/charmbracelet/bubbles/pull/240

PR #328, opened 2025-01-14
fix(ansi): update truncateLeft description, fix edge case

pranavkrishnan02:
Changes:

- Fixed `truncateLeft` description based on @caarlos0 's [PR](https://github.com/charmbracelet/x/pull/301)
- Fixed edge case: Return string if we're truncating 0 chars from the left (see test case) 

Issue: #329

### DIFF
--- a/ansi/truncate.go
+++ b/ansi/truncate.go
@@ -122,13 +122,13 @@ func Truncate(s string, length int, tail string) string {
 	return buf.String()
 }
 
-// TruncateLeft truncates a string from the left side to a given length, adding
-// a prefix to the beginning if the string is longer than the given length.
+// TruncateLeft truncates a string from the left side by removing n characters,
+// adding a prefix to the beginning if the string is longer than n.
 // This function is aware of ANSI escape codes and will not break them, and
 // accounts for wide-characters (such as East-Asian characters and emojis).
-func TruncateLeft(s string, length int, prefix string) string {
-	if length == 0 {
-		return ""
+func TruncateLeft(s string, n int, prefix string) string {
+	if n <= 0 {
+		return s
 	}
 
 	var cluster []byte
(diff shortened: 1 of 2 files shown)


ansi/truncate.go
```diff
@@ -122,13 +122,13 @@ func Truncate(s string, length int, tail string) string {
 	return buf.String()
 }
 
-// TruncateLeft truncates a string from the left side to a given length, adding
-// a prefix to the beginning if the string is longer than the given length.
+// TruncateLeft truncates a string from the left side by a given number of characters, adding
+// a prefix to the beginning if the string is longer than the given number of characters.
 // This function is aware of ANSI escape codes and will not break them, and
 // accounts for wide-characters (such as East-Asian characters and emojis).
```
caarlos0, line 128: Reading this again, it still is a bit confusing I think.

Maybe something like:

```
TruncateLeft returns the last n characters [length] of the given string, 
adding a prefix to the beginning if the given string is longer than [length].
This function is aware of ANSI escape sequences and will not break them.
It also accounts for wide-characters (such as East-Asian characters and emojis).
```

I would also leave the variable named length I think, as to conform with `Truncate`

pranavkrishnan02: I think using `length` is confusing because the current implementations of `Truncate` and `TruncateLeft` aren't complimentary:

`Truncate(s string, length int, "")` returns the first `length` characters of s
`TruncateLeft(s string, length int, "")` returns the last `StringWidth(s) - length` characters of s

Making them complimentary would be ideal but it would break backward compatibility. It would probably look something like:

```
// Returns last [length] characters of s
func TruncateLeft(s string, length int, prefix string) string {
	charsToRemove := StringWidth(s) - length
	if charsToRemove <= 0 {
		return s
	}
        ...
}
```

caarlos0: right, truncate left was added to be used in `Cut`, mostly so we do things like horizontal scrolling. 


anyway its fine.

pranavkrishnan02: ok, tried to make it a bit more clear, should be good now

caarlos0 (approved):

aymanbagabas (approved):

caarlos0:
Thanks @pranavkrishnan02 🙏

PR #338, opened 2025-01-22
Fix handling for `ED (Erase in Display)` control sequence

ryan-gang:
The changes introduced in this PR are:

1. In `vt/csi_screen.go`, the `ED (Erase in Display)` command `case 0` implementation has been updated to handle erasing from the cursor position more precisely.
  The key changes are:
    - Now properly handles erasing from the cursor position by splitting into two rectangles
    - First rectangle erases from cursor to end of current line
    - Second rectangle erases all lines below current line

2. In `vt/terminal_test.go`, all test cases have been updated to use `\x1b[2J` (clear entire screen) instead of `\x1b[0J` (clear from cursor down) for initialization. 

3. `ED` command test cases have also been updated to reflect the updated implementation. 

Ref: [Wikipedia CSI codes](https://en.wikipedia.org/wiki/ANSI_escape_code#Control_Sequence_Introducer_commands), [VT100 ED docs](https://vt100.net/docs/vt510-rm/ED.html)

### DIFF
--- a/vt/csi_screen.go
+++ b/vt/csi_screen.go
@@ -6,15 +6,18 @@ import (
 
 func (t *Terminal) handleScreen() {
 	width, height := t.Width(), t.Height()
-	_, y := t.scr.CursorPosition()
+	x, y := t.scr.CursorPosition()
 
 	switch t.parser.Cmd() {
 	case 'J': // Erase in Display [ansi.ED]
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

aymanbagabas (approved):
Thank you @ryan-gang!

PR #353, opened 2025-02-03
chore(editor): use ints instead of uints to avoid conversions

meowgorithm:
Personal preference, but I find that with `int` as the default numeric unit in Go I'm always a little thrown off when converting to a `uint` for high level cases like this.


caarlos0:
yes, should have been an int all along! thank you

Issue #351, opened 2025-01-31
bug(ansi): truncate includes newlines outside of declared max width


bashbunni:
### Describe the bug

When using text that occupies more than one rune, truncation doesn't ignore newlines in the original input despite them being at an index exceeding the width that should be truncated.

### Setup

Please complete the following information along with version numbers, if applicable.
 - OS [e.g. Ubuntu, macOS] macOS
 - Shell [e.g. zsh, fish] fish
 - Terminal Emulator [e.g. kitty, iterm] kitty
 - Terminal Multiplexer [e.g. tmux] n/a

### To Reproduce

1. apply the patch locally, then run `go test -run TestTruncate -v`

```bash
=== RUN   TestTruncate/multi-width_graphemes_with_newlines_-_japanese_text
    truncate_test.go:298: test case 34 failed: expected "耐許ヱヨカハ…", got "耐許ヱヨカハ…\n\n"
```

Those newlines are from way further down, beyond the width of 14 that I've given this test case. I stepped through it in a debugger and found that when i = 535 in `Truncate`, it hits the default case and writes that cell despite all the other surrounding values being ignored.

![Image](https://github.com/user-attachments/assets/dfea5f48-18f8-44bd-90ae-6189363b6787)

## Patch

```bash
diff --git a/ansi/truncate_test.go b/ansi/truncate_test.go
index 0f346df..e104b1c 100644
--- a/ansi/truncate_test.go
+++ b/ansi/truncate_test.go
@@ -277,6 +277,18 @@ var tcases = []struct {
                "สวัสดีสวัสดี\x1b]8;;https://example.com\x1b\\\n…\x1b]8;;\x1b\\",
                "\x1b]8;;https://example.com\x1b\\\n…วัสดีสวัสดี\x1b]8;;\x1b\\",
        },
+       {
+               "multi-width graphemes with newlines - japanese text",
[...]

PR #367, opened 2025-02-10
fix(ansi): cut/truncate/truncateleft improperly handling new lines

caarlos0:
I'm not sure this is the best way to handle this, but, without this
patch, truncates would still use new lines from outside its cut target.

This fixes it.

closes https://github.com/charmbracelet/x/issues/351


aymanbagabas:
This is related to https://github.com/charmbracelet/x/issues/351

aymanbagabas (approved):

aymanbagabas (commented):
TruncateLeft still ignores newlines even if they're within the range

caarlos0:
> TruncateLeft still ignores newlines even if they're within the range

seems to be working fine to me, even added another test for it.

PS: the problem was that it **wasn't** ignoring out of the cut new lines - so it was including all of them.

ansi/truncate.go
```diff
@@ -142,6 +143,14 @@ func truncate(m Method, s string, length int, tail string) string {
 			// collects printable ASCII
 			curWidth++
 			fallthrough
+		case parser.ExecuteAction:
```
raphamorim, line 146: so in case of a new line, it will be ignored?

raphamorim (approved):
👍

ansi/truncate.go
raphamorim: nevermind saw your comment

aymanbagabas (approved):

PR #380, opened 2025-02-19
feat: add sixel to ansi package

raphamorim:
PR originally implemented by @CannibalVox

Ref: https://github.com/charmbracelet/x/pull/352

Most of this PR changes are just updates on the implementation per requests.

## API

Here's what the API looks like.

```go
import  "github.com/charmbracelet/x/ansi/sixel"

var (
    buf bytes.Buffer
    e sixel.Encoder
    img image.Image // imagine this is populated with image data
)


if err := e.Encode(&buf, img); err != nil {
	return 0, err
}

_, _ = io.WriteString(os.Stdout, ansi.SixelGraphics(0, 1, 0, buf.Bytes()))
```

## Example

`cd examples/img2term && go run . ./../../ansi/fixtures/graphics/JigokudaniMonkeyPark.png`

Screenshot 2025-02-21 at 14 21 22

## Misc

Example while running go bench with `fmt.Println` iterm2/wezterm/rio:

image


ansi/graphics.go
```diff
@@ -8,11 +8,65 @@ import (
...
+	buf.Write(payload)
+	buf.WriteString("\x1b\\")
+
+	return buf.String()
+}
+
+// WriteSixelGraphics encodes an image as sixels into the provided io.Writer.
+// Options is provided in expectation of future options (such as dithering), but
+// none are yet implemented. o can be nil to use the default options.
+func WriteSixelGraphics(w io.Writer, m image.Image) error {
```
aymanbagabas, line 58: I think "Encode" signifies the meaning more

```suggestion
// EncodeSixelGraphics encodes an image as sixels into the provided io.Writer.
// Options is provided in expectation of future options (such as dithering), but
// none are yet implemented. o can be nil to use the default options.
func EncodeSixelGraphics(w io.Writer, m image.Image) error {
```

ansi/graphics.go
```diff
@@ -8,11 +8,65 @@ import (
...
+	}
+	buf.WriteByte(';')
+	if p2 >= 0 {
+		buf.WriteString(strconv.Itoa(p2))
+	}
+	buf.WriteByte(';')
+	if p3 >= 0 {
+		buf.WriteString(strconv.Itoa(p3))
+	}
+	buf.WriteString(";q")
```
aymanbagabas, line 48:
```suggestion
	buf.WriteString("q")
```

ansi/graphics.go
```diff
@@ -8,11 +8,65 @@ import (
...
+		buf.WriteString(strconv.Itoa(p1))
+	}
+	buf.WriteByte(';')
+	if p2 >= 0 {
+		buf.WriteString(strconv.Itoa(p2))
+	}
+	buf.WriteByte(';')
+	if p3 >= 0 {
+		buf.WriteString(strconv.Itoa(p3))
+	}
```
aymanbagabas, line 47: Since p3 is not really used by terminal emulators, we can safely ignore it if it's zero.
```suggestion
	if p3 > 0 {
		buf.WriteByte(';')
		buf.WriteString(strconv.Itoa(p3))
	}
```

ansi/graphics.go
aymanbagabas: I wonder if we actually need this function 🤔

Implementors can use the `sixel` package and write the encoded output in a `ansi.SixelGraphics`. This way they can encode the images with whatever encoding options we have.

raphamorim: yea I would rather remove this function

ansi/sixel/encoder.go
```diff
@@ -0,0 +1,270 @@
...
+
+type Options struct {
+}
+
+type Encoder struct {
+}
+
+func Raster(pan, pad, ph, pv int) string {
+	return fmt.Sprintf("%s%d;%d;%d;%d", string(RasterAttribute), pan, pad, ph, pv)
+}
```
aymanbagabas, line 40: Perhaps we can have a `type Raster struct { Pan, Pad, Ph, Pv int}` with `func (Raster) Encode(io.Writer) error`, `func (Raster) String() string` that uses `r.Encode`, and `func ParseRaster([]byte) (Raster, error)`. This can go into the `Encoder` struct 

Then, if we have a `Pad == 0`, we default to `Raster{1, 1, img.Dx(), img.Dy}`

ansi/sixel/encoder.go
```diff
@@ -0,0 +1,270 @@
...
+	RepeatIntroducer byte = '!'
+	ColorIntroducer  byte = '#'
+	RasterAttribute  byte = '"'
+)
+
+type Options struct {
+}
+
+type Encoder struct {
+}
```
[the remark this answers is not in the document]
raphamorim: I agree, but I think this is a good follow up change, what do you think?

ansi/sixel/encoder.go
```diff
@@ -0,0 +1,270 @@
...
+	return fmt.Sprintf("%s%d;%d;%d;%d", string(RasterAttribute), pan, pad, ph, pv)
+}
+
+func Repeat(count int, repeatRune rune) string {
+	var sb strings.Builder
+	sb.WriteByte(RepeatIntroducer)
+	sb.WriteString(strconv.Itoa(count))
+	sb.WriteRune(repeatRune)
+	return sb.String()
+}
```
aymanbagabas, line 48: Same, I think a `type Repeat struct { Count int; Char byte }` makes more sense

ansi/graphics.go
```diff
@@ -8,11 +8,50 @@ import (
...
+
+	buf.WriteString("\x1bP")
+	if p1 >= 0 {
+		buf.WriteString(strconv.Itoa(p1))
+	}
+	buf.WriteByte(';')
+	if p2 >= 0 {
+		buf.WriteString(strconv.Itoa(p2))
+	}
+	buf.WriteByte(';')
```
aymanbagabas, line 43:
```suggestion
```

ansi/sixel/encoder.go
```diff
@@ -0,0 +1,270 @@
...
+type sixelBuilder struct {
+	SixelPalette sixelPalette
+
+	imageHeight int
+	imageWidth  int
+
+	pixelBands bitset.BitSet
+
+	imageData   strings.Builder
+	repeatRune  rune
```
aymanbagabas, line 118:
```suggestion
	repeatChar  byte

```

ansi/sixel/palette.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+// sixelColor is a flat struct that contains a single color: all channels are 0-100
+// instead of anything sensible
+type sixelColor struct {
+	Red   uint32
+	Green uint32
+	Blue  uint32
+	Alpha uint32
+}
```
aymanbagabas, line 268: I wonder if it makes more sense if we have `type Color struct { Pc, Pu uint8; Px, Py Pz float64 }` to reflect the color type defined in the Sixel specs which can be a paletted color, RGB (channels are 0-100), or HLS color (Px would be 0-360, Py and Pz are 0-100)

Then we can have helper functions the same as the other types, `func (Color) Encode(io.Writer) error`, `func (Color) String() string`, and `func ParseColor([]byte) (Color, error)`

ansi/sixel/palette.go
```diff
@@ -0,0 +1,385 @@
...
+		cube.score = uint64(dAlpha)
+	}
+
+	// Boost the score of cubes with more pixels in them
+	cube.score *= totalWeight
+
+	return cube
+}
+
+type Ordered interface {
```
aymanbagabas, line 162: We don't need to export these types. Probably we should move them to `utils.go`
```suggestion
type ordered interface {

```

ansi/sixel/palette.go
```diff
@@ -0,0 +1,385 @@
...
+	~int | ~int8 | ~int16 | ~int32 | ~int64 |
+		~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr |
+		~float32 | ~float64 |
+		~string
+}
+
+
+// isNaN reports whether x is a NaN without requiring the math package.
+// This will always return false if T is not floating-point.
+func isNaN[T Ordered](x T) bool {
```
aymanbagabas, line 172:
```suggestion
func isNaN[T ordered](x T) bool {

```

ansi/sixel/palette.go
```diff
@@ -0,0 +1,385 @@
...
+	return x != x
+}
+
+func Compare[T Ordered](x, y T) int {
```
aymanbagabas, line 176:
```suggestion
func compare[T ordered](x, y T) int {

```

ansi/sixel/color.go
```diff
@@ -13,14 +13,37 @@ var ErrInvalidColor = fmt.Errorf("invalid color")
...
+	return (c + 328) * 100 / 0xffff
+}
+
+// FromColor returns a Sixel color from a color.Color. It converts the color
+// channels to the 0-100 range.
+func FromColor(c color.Color) Color {
+	r, g, b, a := c.RGBA()
+	if a == 0 {
+		return Color{Pu: 3} // Transparent color
+	}
```
aymanbagabas, line 38: We don't need this

ansi/sixel/encoder.go
```diff
@@ -0,0 +1,277 @@
...
+	NumColors int
+
+	// AddTransparent is a flag that indicates whether to add a transparent
+	// color to the palette. The default is false.
+	AddTransparent bool
+
+	// TransparentColor is the color to use for the transparent color in the
+	// palette. If nil, [color.Transparent] will be used.
+	// This field is ignored if [Encoder.AddTransparent] is false.
+	TransparentColor color.Color
```
aymanbagabas, line 46: There options are not used

ansi/sixel/decoder.go
```diff
@@ -0,0 +1,530 @@
...
+			if c.Pu > 0 {
+				// Non-zero Pu means we have a color definition to set.
+				palette[currentPaletteIndex] = c
+			}
+			// palette[currentPaletteIndex] = color.RGBA64{
+			// 	R: uint16(imageConvertChannel(uint32(c.Px))),
+			// 	G: uint16(imageConvertChannel(uint32(c.Py))),
+			// 	B: uint16(imageConvertChannel(uint32(c.Pz))),
+			// 	A: 65525,
+			// }
```
aymanbagabas, line 131:
```suggestion
```

caarlos0 (approved):

raphamorim:
iterm2/wezterm/rio

image

meowgorithm (approved):
I can't speak to the implementation details, but functionally speaking, LGTM

PR #387, opened 2025-02-27
feat: halfblocks support

raphamorim:
Support to half

# API

```go
raw, err := loadImage("./../fixtures/graphics/JigokudaniMonkeyPark.png")
if err != nil {
	os.Exit(1)
}

// Create options
options := EncoderOptions{
	ColorMode:    3,
	Width:        80,
	Height:       0,
	Threshold:    uint8(128),
	DitherLevel:  0.0,
	UseFgBgOnly:  false,
	InvertColors: false,
	ScaleMode:    "default",
	Symbols:      "half",
}

result := Encode(raw, options)
fmt.Println(result)
```

Result:

Screenshot 2025-02-27 at 12 33 37

# Other examples

Screenshot 2025-02-27 at 12 35 31


ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	_ "image/jpeg"
+	_ "image/png"
+	"math"
+	"strings"
+)
+
+// Block represents different Unicode block characters
+type Block struct {
+	Char        rune
+	Coverage    []bool // Which parts of the block are filled (true = filled)
```
aymanbagabas, line 17: This is always gonna be 4 sides
```suggestion
	Coverage    [4]bool // Which parts of the block are filled (true = filled)
```

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	CoverageMap string // Visual representation of coverage for debugging
+}
+
+// EncoderOptions contains all configurable settings
+type EncoderOptions struct {
```
aymanbagabas, line 22:
```suggestion
type Options struct {
```

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	ColorMode    int     // 0=none, 1=8colors, 2=256colors, 3=truecolor
```
aymanbagabas, line 23: we probably don't need this since downsampling colors is handled by `colorprofile`

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	Width        int     // Output width
+	Height       int     // Output height (0 for auto)
+	Threshold    uint8   // Threshold for considering a pixel as set (0-255)
+	DitherLevel  float64 // Dithering amount (0.0-1.0)
+	UseFgBgOnly  bool    // Use only foreground/background colors (no block symbols)
+	InvertColors bool    // Invert colors
+	ScaleMode    string  // fit, stretch, center
```
aymanbagabas, line 30: Let's define a type for this one, maybe use `draw.Scaler` https://pkg.go.dev/golang.org/x/image/draw#Scaler

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	Symbols      string  // Which symbols to use: "half", "quarter", "all"
```
aymanbagabas, line 31: Same, I think this should be a type `type Symbol string`

raphamorim:
btw we use the term enconding but in halfblocks isn't reaaaallly encoding. I think makes more sense to rename to render term

raphamorim:
talked with @aymanbagabas and we think it makes more sense to live in lipgloss. Will be moving it to there and address the comments

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
+package halfblocks
+
+import (
+	"fmt"
+	"image"
+	"image/color"
+	_ "image/gif"
+	_ "image/jpeg"
+	_ "image/png"
```
aymanbagabas, line 9: We shouldn't import image encodings in libraries, this should happen in the `main` package, and it is the responsibility of the user

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+		DitherLevel:  0.0,       // Medium dithering
+		UseFgBgOnly:  false,     // Use block symbols
+		InvertColors: false,     // Don't invert
+		ScaleMode:    "default", // Fit to terminal
+		Symbols:      "half",    // Use half blocks
+	}
+}
+
+// GenerateBlocks returns supported block characters based on the symbols option
+func GenerateBlocks(symbolsOption string) []Block {
```
aymanbagabas, line 60: We can have defined as globals

```go
var (
    HalfBlocks = []Block{...}
    QuarterBlocks = []Block{...}
    ...
)
```

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	case 3: // True color
+		fgStr = fmt.Sprintf("\033[38;2;%d;%d;%dm", fg.R, fg.G, fg.B)
+		bgStr = fmt.Sprintf("\033[48;2;%d;%d;%dm", bg.R, bg.G, bg.B)
+	}
+
+	return fgStr + bgStr + string(char) + "\033[0m"
+}
+
+// getPixelSafe returns the color at (x,y) or black if out of bounds
+func (r *Renderer) getPixelSafe(img image.Image, x, y int) color.RGBA {
```
aymanbagabas, line 337: Let's use `color.Color` instead of `color.RGBA` all over

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	bounds := img.Bounds()
+	if x < bounds.Min.X || x >= bounds.Max.X || y < bounds.Min.Y || y >= bounds.Max.Y {
+		return color.RGBA{0, 0, 0, 255}
```
aymanbagabas, line 340: This can be
```suggestion
		return color.Black
```

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+	}
+
+	r8, g8, b8, a8 := img.At(x, y).RGBA()
+	return color.RGBA{
+		R: uint8(r8 >> 8),
+		G: uint8(g8 >> 8),
+		B: uint8(b8 >> 8),
+		A: uint8(a8 >> 8),
+	}
```
aymanbagabas, line 349:
```suggestion
	return img.At(x, y)
```

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+}
+
+// scaleImage resizes an image to the specified dimensions
+func (r *Renderer) scaleImage(img image.Image, width, height int) image.Image {
```
aymanbagabas, line 353: All scale functions should adhere to [`draw.Scaler`](https://pkg.go.dev/golang.org/x/image/draw#Scaler)

ansi/halfblocks/encoder.go
```diff
@@ -0,0 +1,626 @@
...
+		bgCode := nearestAnsi256Color(bg.R, bg.G, bg.B)
+		fgStr = fmt.Sprintf("\033[38;5;%dm", fgCode)
+		bgStr = fmt.Sprintf("\033[48;5;%dm", bgCode)
+
+	case 3: // True color
+		fgStr = fmt.Sprintf("\033[38;2;%d;%d;%dm", fg.R, fg.G, fg.B)
+		bgStr = fmt.Sprintf("\033[48;2;%d;%d;%dm", bg.R, bg.G, bg.B)
+	}
+
+	return fgStr + bgStr + string(char) + "\033[0m"
```
aymanbagabas, line 333: We should use `ansi.Style` with `color.Color` instead. Downgrading colors can should be handled by colorprofile. We have types for 16-bit and indexed in `ansi`, specifically `ansi.BasicColor`, `ansi.IndexedColor`, any other `color.Color` type will be treated as a true color.

The only optimization I can think of is checking if the RGB value corresponds to a 256 terminal color and using that to encode the color saving some SGR bytes i.e. if the color is `#5f0000` we can use the defined XTerm dark red (indexed 52) color instead of to save bytes `\x1b[48;2;95;0;0m` vs `\x1b[48;5;52m'.

Anyway, we should use `ansi.Style{}.ForegroundColor(fg).BackgroundColor(bg).Styled(string(char))` instead

ansi/halfblocks/encoder.go
raphamorim: Addressed

ansi/halfblocks/encoder.go
raphamorim: you are right, I forgot to remove it thx!

PR #407, opened 2025-03-12
fix(cellbuf): wrap: properly handle formatted spaces

aymanbagabas:
This will treat formatted styled spaces as word characters. We need this because for example a space with a background shouldn't be truncated by the wrap.

### DIFF
--- a/cellbuf/cell.go
+++ b/cellbuf/cell.go
@@ -154,6 +154,11 @@ const (
 	ResetAttr AttrMask = 0
 )
 
+// Contains returns whether the attribute mask contains the attribute.
+func (a AttrMask) Contains(attr AttrMask) bool {
+	return a&attr == attr
+}
+
 // UnderlineStyle is the style of underline to use for text.
 type UnderlineStyle = ansi.UnderlineStyle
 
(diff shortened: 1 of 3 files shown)


andreynering (commented):
Hey @aymanbagabas, take a look at these test cases that I wrote:

```go
{
	name:     "long text inside color formatting",
	input:    "\x1b[38;5;203;48;5;236marm64 amd64 i386\x1b[0m",
	expected: "\x1b[38;5;203;48;5;236marm64 \x1b[0m\n\x1b[38;5;203;48;5;236mamd64 \x1b[0m\n\x1b[38;5;203;48;5;236mi386 \x1b[0m",
	width:    10,
},
{
	name:     "long text inside bold formatting",
	input:    "\x1b[1marm64 amd64 i386\x1b[0m",
	expected: "\x1b[1marm64\x1b[m\n\x1b[1mamd64\x1b[m\n\x1b[1mi386\x1b[0m",
	width:    7,
},
```

Screenshot 2025-03-12 at 14 06 42

For a simple formatting as bold it is working fine, but for color formatting it is now not breaking when it could.

That looks expected considering what the code is doing, but we might want to address that, if possible.

PR #414, opened 2025-03-27
feat(ansi): add x11 color names for osc 10-12

mkotowski:
As per xterm's ctlseqs OSC 10-19 colors:

> [...] [are specified by name or RGB specification as per XParseColor](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#:~:text=The%20colors%20are%20specified%20by%20name%20or%20RGB%20specification%0A%20%20%20%20%20%20%20%20%20%20as%20per%20XParseColor.)

This commit extends x/ansi `XParseColor` implementation to also support the X11 color names.

Optionally, `vt`'s `handleDefaultColor` can also be extended to ignore strings longer than the max length of the passed string (65535u to match xterm's behavior with `MAX_U_STRING` const).


aymanbagabas:
Thank you, @mkotowski, for sending this PR. Why don't you simply use [`image/colornames`](https://pkg.go.dev/golang.org/x/image/colornames) instead?

PR #420, opened 2025-03-31
docs(cellbuf): add usage example to cellbuf.Wrap

bashbunni:
optional, but I thought this might be helpful to new users since we're using it more widely across projects now

image


meowgorithm:
Per @andreynering, let's merge this since it's simple and already done.

PR #429, opened 2025-04-21
feat(scripts): add scripts/lints

aymanbagabas:
This generates a lint workflow for each go module that we have under `/x`. It uses our shared lint workflow from
`meta/.github/workflows/lint.yml`.


scripts/lints
```diff
@@ -0,0 +1,32 @@
...
+  # get the parent directory when the module is nested semver-style
+  if [[ "$name" =~ ^v[0-9]+.*$ ]]; then
+    ver="$(basename "$dir")"
+    dir="$(dirname "$dir")"
+    name="$(basename "$dir")-$ver"
+  fi
+
+  moddir="$(dirname "$mod")"
+
+  echo "# auto-generated by scripts/builds. DO NOT EDIT.
```
caarlos0, line 16:
```suggestion
  echo "# auto-generated by scripts/lints. DO NOT EDIT.
```

could also do both lint and build workflows in the ./scripts/builds file 🤔

caarlos0 (approved):

scripts/lints
aymanbagabas: Yeah, less code to maintain

caarlos0 (approved):

PR #430, opened 2025-04-21
chore(exp/charmtone): @aymanbagabas fixes the API

meowgorithm:


exp/charmtone/charmtone.go
```diff
@@ -187,7 +193,7 @@ func Hexes() map[Key]string {
 		Smoke:    "#BFBCC8", // Gray 5
 		Ash:      "#DFDBDD", // Gray 6
 		Butter:   "#FFFAF1",
-	}
+	}[k]
 }
 
 func Keys() []Key {
```
aymanbagabas, line 199: Perhaps rename this into `func Colors() []Key` 🤔

PR #443, opened 2025-04-29
feat(exp/slice): add `slice.Last` helper func

andreynering:
Just a small convenience function to have clearer code.

```go
last := obj.List[len(obj.List)-1]
```

Becomes:

```go
last := slice.Last(obj.List)
```

### DIFF
--- a/exp/slice/slice.go
+++ b/exp/slice/slice.go
@@ -22,14 +22,14 @@ func Take[A any](slice []A, n int) []A {
 	return slice[:n]
 }
 
-// Last returns the last element of a slice. If the slice is empty, it
-// returns the zero value.
-func Last[T any](list []T) T {
+// Last returns the last element of a slice and true. If the slice is empty, it
+// returns the zero value and false.
+func Last[T any](list []T) (T, bool) {
 	if len(list) == 0 {
 		var zero T
-		return zero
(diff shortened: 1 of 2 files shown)


exp/slice/slice.go
```diff
@@ -21,6 +22,16 @@ func Take[A any](slice []A, n int) []A {
 	return slice[:n]
 }
 
+// Last returns the last element of a slice. If the slice is empty, it
+// returns the zero value.
+func Last[T any](list []T) T {
```
caarlos0, line 27: I think I'll do something like:

```go
func Last[T any](list []T) (T, bool) {
	if len(list) == 0 {
		var zero T
		return zero, true
	}
	return list[len(list)-1], false
```

because its probably useful to know when the result is the zero value because that was the value in the slice vs when the slice is actually empty

andreynering: I considered that, and initially decided against because it makes it less convenient to use directly on scenarios like this:

```go
doSomething(slice.Last(list))
```

That now will have to be written as:

```go
item, ok := slice.Last(list)
if ok {
        doSomething(item)
}
```

But I'm okay with it, so I proceeded with the change.

aymanbagabas (approved):

PR #463, opened 2025-05-12
chore(test): tidy mosaic test

bashbunni:
The test was failing, but giving no indication why, so I cleaned up some of the logs in the test here for debugging


meowgorithm (changes requested):
This looks good. Would also resolve the linter suggestions as well: there are only a few.

bashbunni:
ready 2 rock n' roll for rapha

PR #478, opened 2025-06-02
feat(exp/golden): accept strings on `RequireEqual`

andreynering:
With generics we can allow both `[]byte` and `string`, so no more `[]byte(foo)` casting needed from now on.


exp/golden/golden.go
```diff
@@ -55,7 +55,7 @@ func RequireEqual(tb testing.TB, out []byte) {
 // the expected from the golden files, printing its diff in case it is not.
 //
 // Deprecated: Use [RequireEqual] instead.
-func RequireEqualEscape(tb testing.TB, out []byte, escapes bool) { //nolint:revive
+func RequireEqualEscape[T []byte | string](tb testing.TB, out T, escapes bool) { //nolint:revive
 	RequireEqual(tb, out)
 }
```
ccoVeille, line 60: This one is deprecated, I don't think there is a need to update it

aymanbagabas (approved):

caarlos0 (approved):

PR #496, opened 2025-07-01
ci: make lint actions green

andreynering:
This PR fixes many lint issues spread across many of `x` packages. They started to report once we updated the rules to be the same for all our repos.

Notes:

* `vt` and `teatest-v2` builds were already broken on `main`. I left as is in this PR.
* `windows` and `input` packages have issues that don't report locally on non-Windows machines

TODO:

* [ ] Fix `windows` package issues
* [ ] Fix `input` package issues


caarlos0:
fwiw #471

caarlos0 (approved):

PR #499, opened 2025-07-07
 fix: incorrect parsing on WSL due to buffering of osc esc sequences

rgodha24:
it passes all tests and fixes the repro described in the bubbletea issue (first link below). 

related: 
- https://github.com/charmbracelet/bubbletea/issues/1440
- https://github.com/charmbracelet/bubbletea/issues/1406
- https://github.com/charmbracelet/x/pull/497
- https://github.com/charmbracelet/x/commit/39faa120bfa597a373e044bd1ebe8f36d2d822f4

NOTE: this is completely vibe coded. I would not get mad at all if this is closed bc of subpar code quality


rgodha24:
unfortunately i can only test with windows terminal because of work computer restrictions.

from my very limited googling, it seems like wezterm does not support the osc 11 query, but this could be incorrect.

lrstanley:
> unfortunately i can only test with windows terminal because of work computer restrictions.
> 
> EDIT: the below probably isnt true. one of the original issues that led me down this rabbithole mentioned they were having this issue on Wezterm as well ([sst/opencode#127 (comment)](https://github.com/sst/opencode/issues/127#issuecomment-2978440226)). @lrstanley would you be able to test the repro in the first issue linked in the description on wezterm?
> 
> from my very limited googling, it seems like wezterm does not support the osc 11 query, but this could be incorrect.

Not sure which repro you're referring to, unless you mean running opencode itself on that specific commit (before the patch was implemented?)

I did see https://github.com/wezterm/wezterm/issues/5899 -- and although this is supported through ConPTY now I think, wezterm hasn't released in a while and might depend on something older with such an old build.

Unless there is a usecase where we know it should return a response that I can test with, I'd say it's probably fine to continue with this fix, unless someone else notices specific issues.

lrstanley:
I think it's pending primarily due to the new https://github.com/charmbracelet/ultraviolet package, that I think they're about to switch to using in the v2-exp branches of bubbletea/lipgloss. I haven't had time to confirm yet if it solves the issues though.

lrstanley:
FYI, I think the issue still occurs with the ultraviolet implementation, due to the same issues outlined here (the escape sequences caught at the end of 1 buffer don't get cleanly grouped into the next received buffer). Working with Ayman on the issue.

PR #518, opened 2025-08-09
perf(cellbuf): reuse ansi.Parser instances in PenWriter

tazjin:
The parsers used in PenWriter have very large internal buffers, but were created from scratch for each writer.

In crush this led to a lot of allocation churn, which had a noticeable performance impact.

The fix is to use a `sync.Pool` (similar to `ansi.GetParser`), but separate from it, as the parser parameters are different.

Since this is stored *in* the type, it must be released afterwards. For this reason the `Close` method is extended appropriately, and must (in a subsequent commit to `glamour`) be used everywhere.

Note that **not** calling `Close` is not a correctness issue: The program will still run correctly, no parsers will leak, it'll just be slower.

Due to the required changes in `glamour`, this change *on its own* has no performance impact, but is a necessary prerequesite for the impactful change.

### Checklist before requesting a review

- [ ] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my code

### If this is a feature

- [ ] I have created a discussion
- [ ] A project maintainer has approved this feature request. Link to comment:


tazjin:
@aymanbagabas Thanks for the review! It is important that the dependency is now updated in `glamour`, so that the fix over there (https://github.com/charmbracelet/glamour/pull/465) starts working.

PR #532, opened 2025-08-12
feat(ansi): add linux console palette colors sequences

aymanbagabas:
This adds support for setting the color palette in the Linux Console using OSC P and OSC R sequences. The `SetPalette` function allows setting a specific color for a given index (0-15), and the `ResetPalette` constant resets the palette to default values.


tauraamui (commented):
I know it's just doing a single string generation and we can't actually validate the behaviour of the escape code this way, but it might be nice to have a unit test or two.

meowgorithm (approved):

PR #540, opened 2025-08-28
feat(exp/slice): add `slice.IsSubset`

andreynering:
This was added temporarily to Crush. Once merged here, we can point Crush to this package to use this function.

Please, let me know if you have a better suggestion as a name for the function.


meowgorithm (approved):
I love it when you add stuff like this to the x repo!

PR #632, opened 2025-11-09
fix(cellbuf): update ansi, colorprofile

caarlos0:
fixes build issues when using latest lipgloss as well.

refs https://github.com/charmbracelet/fang/issues/80


caarlos0:
hah, so we're missing tagging lipgloss probably

raphamorim (approved):

PR #652, opened 2025-11-18
feat(vt): add damage callbacks for terminal updates

aymanbagabas:
This commit introduces a new Damage callback in the vt package that notifies when regions of the terminal are modified. This allows clients to efficiently track changes and update their displays accordingly.

### DIFF
--- a/vt/utf8.go
+++ b/vt/utf8.go
@@ -9,10 +9,18 @@ import (
 
 // handlePrint handles printable characters.
 func (e *Emulator) handlePrint(r rune) {
+	e.handleRune(r, true)
+}
+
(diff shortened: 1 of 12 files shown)


vt/utf8.go
```diff
@@ -22,7 +30,7 @@ func (e *Emulator) handlePrint(r rune) {
 
 // flushGrapheme flushes the current grapheme buffer, if any, and handles the
 // grapheme as a single unit.
-func (e *Emulator) flushGrapheme() {
+func (e *Emulator) flushGrapheme(damage bool) {
```
raphamorim: (hasDamage bool)

PR #653, opened 2025-11-20
feat(tone): remove dependencies on go-colorful and exp/color

aymanbagabas:
This simply removes the dependencies by hardcoding the color values as RGBA structs and removing the Blend function as it's now part of lipgloss.

### DIFF
--- a/exp/charmtone/charmtone_test.go
+++ b/exp/charmtone/charmtone_test.go
@@ -1,16 +1,19 @@
 package charmtone
 
 import (
-	"regexp"
+	"strconv"
+	"strings"
 	"testing"
 )
 
-var hexRegexp = regexp.MustCompile(`^#(?:[0-9a-fA-F]{3}){1,2}$`)
-
 func TestValidateHexes(t *testing.T) {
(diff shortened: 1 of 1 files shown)


exp/charmtone/charmtone_test.go
```diff
@@ -1,15 +1,16 @@
...
+
 func TestValidateHexes(t *testing.T) {
 	for _, key := range Keys() {
-		if _, err := colorful.Hex(key.Hex()); err != nil {
-			t.Errorf("Key %s: %v", key, err)
+		if !hexRegexp.MatchString(key.Hex()) {
+			t.Errorf("Key %s: invalid hex format %s", key, key.Hex())
 		}
 	}
 }
```
ccoVeille: What about using something like this

```go
h, ok := strings.CutPrefix(.Hex())
if !ok || len(h) != 6
   t.Fatal

_, err := strings.ParseUint(h, 16)
if err != nil
   t.Fatal

Regexp are slow, and here you only want to validate an hex code

PR #790, opened 2026-03-02
fix(powernap): use req.Notif to detect notifications in Router

lexh:
`Router.Route` compared `req.ID` to `jsonrpc2.ID{}` to distinguish notifications from requests. Since `jsonrpc2.ID{}` has `Num: 0`, any server-initiated request with integer ID 0 (e.g. the first outgoing request from lsp_server/ty) was misrouted as a notification, skipping its registered handler entirely.

Use `req.Notif` instead, which is correctly set by `Request.UnmarshalJSON` based on whether the "id" field is present in the JSON.

Also guard against nil `req.Params` to prevent panics when a server sends a request or notification without a params field.

- [ ] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/powernap/pkg/transport/router.go
+++ b/powernap/pkg/transport/router.go
@@ -57,22 +57,30 @@ func (r *Router) Route(ctx context.Context, req *jsonrpc2.Request) (any, error) 
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 
-	// Check if it's a notification (no ID)
-	if req.ID == (jsonrpc2.ID{}) {
+	// Safely extract params, defaulting to "null" if not present.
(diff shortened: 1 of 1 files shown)


squarepear:
Not a maintainer, so I don't want to leave an official approval, but the change seems reasonable.

I was running into the same issue when using [`ty`](https://docs.astral.sh/ty/) as an LSP with Crush. I tested this PR rebased on latest and overriding crush's go.mod with my local clone. The issue seems to be fully resolved with this change and Crush can properly use symbols, definitions, references, and call hierarchy with `ty`!

Before:
image

After:
image

PR #850, opened 2026-04-29
fix(vt): prevent data race between Read/Write/Close in SafeEmulator

lrstanley:
## Changes in this PR

- Fixes a data race, due to `SafeEmulator.Close()` (a write) falling through to `Emulator`, where it's also read by `Read()` and `Write()`
- `Read()` also wasn't locking in any form, which was problematic due to the race being with `Emulator.closed` in particular. Can't block the entire operation, since it looks like the pipe reader (`pr`) also blocks.

Only other alternative I could think of would be to use an `atomic.Bool` on `Emulator.closed`, but that bleeds concurrency protection into `Emulator`.

---

### DIFF
--- a/vt/safe_emulator.go
+++ b/vt/safe_emulator.go
@@ -2,6 +2,7 @@ package vt
 
 import (
 	"image/color"
+	"io"
 	"sync"
 
 	uv "github.com/charmbracelet/ultraviolet"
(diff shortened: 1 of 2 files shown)


aymanbagabas (approved):

aymanbagabas:
Thank you @lrstanley!

PR #853, opened 2026-05-06
fix(powernap): handle leading slash in Windows file URI paths

sven2718:
I've been running into issues with LSPs on windows in Crush.  This small PR would fix the LSP/windows bugs I'm seeing atm, I think.

As best as I can tell, the basic problem is that the existing code does not account for the fact that the windows drive letter will usually be *prefixed* by a `/`, so the current logic isn't working as intended, because it sees a path that starts with `/`, and assumes it's not actually dealing with a windows path.

Opus-4.7 wrote a little fix for me.  It also added protocol/uri_test.go (the package previously had no test file). Cases cover empty, POSIX absolute, POSIX with %20, plus four Windows-drive shapes that real LSP clients send: bare drive, lowercase drive (must uppercase), percent-encoded colon (`C%3A`), and a path with a space. Without this fix, all four Windows-drive cases fail with a leading `\` on the result.

💘 Generated with Crush

Assisted-by: Claude Opus 4.7 via Crush <crush@charm.land>

- [ NA ] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/powernap/pkg/lsp/protocol/uri.go
+++ b/powernap/pkg/lsp/protocol/uri.go
@@ -114,10 +114,13 @@ slow:
 	if u.Scheme != fileScheme {
 		return "", fmt.Errorf("only file URIs are supported, got %q from %q", u.Scheme, uri)
 	}
-	// If the URI is a Windows URI, we trim the leading "/" and uppercase
-	// the drive letter, which will never be case sensitive.
-	if isWindowsDrivePath(u.Path) {
-		u.Path = strings.ToUpper(string(u.Path[1])) + u.Path[2:]
+	// If the URI is a Windows URI, trim the leading "/" and uppercase
+	// the drive letter (which is never case sensitive). url.URL.Path
+	// stores Windows file URIs in /C:/foo form, with a leading slash
+	// that hides the drive letter from isWindowsDrivePath; trim it
+	// before consulting the helper.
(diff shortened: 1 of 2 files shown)


taciturnaxolotl (approved):

taciturnaxolotl:
that looks great! thank you!

taciturnaxolotl:
lint and build are failing atm

sven2718:
qwen-3.6 seems to think the issue is that your linting setup is a little too strict.  i'm not certain i trust it on that?  let me see if i can find a less-invasive fix.

sven2718:
yeah, qwen was chasing other preexisting lint issues.  force pushed something simpler that i'm more confident should pass CI.

taciturnaxolotl:
build is failing now 😭

sven2718:
> build is failing now 😭

i'm pretty sure a lot of these lint failures are not my fault 😉 they look prexisting.  that said, opus-4.7 thinks it can fix them all with a relatively focused pr.  
https://github.com/charmbracelet/x/pull/854
you could run the CI there and see if it's right 🙂

taciturnaxolotl:
can you rebase onto main for your ci fix?

sven2718:
> can you rebase onto main for your ci fix?

sure.  seems like ci is happier now 🙂

PR #867, opened 2026-05-18
chore: update charmtone

meowgorithm:
This updates the Charmtone library and its the associated CLI utility. Some color names have changed and have been deprecated. Note that as this library is not stable color names (deprecated and otherwise) can be removed at any time.

image


meowgorithm:
One of the examples is failing due to a timing issue. It was broken prior to this PR. Merging to keep things moving; we can fix that issue in a separate pass.

PR #881, opened 2026-06-01
fix(vt): use atomic.Bool for Emulator.closed to prevent data race

### LINKED ISSUE #879: vt: data race on Emulator.closed between Read and Close
`Emulator.Read` and `Emulator.Close` both touch the unexported `closed bool` with no synchronization, and `SafeEmulator` doesn't protect it either — so calling `Close()` to unblock a goroutine parked in `Read()` is a data race.

`Emulator.Read` reads `closed`, then blocks on the pipe (`vt/emulator.go:251`):

```go
func (e *Emulator) Read(p []byte) (n int, err error) {
	if e.closed {        // unsynchronized read
		return 0, io.EOF
	}
	return e.pr.Read(p)  // blocks here
}
```

`Emulator.Close` writes it (`vt/emulator.go:260`):

```go
func (e *Emulator) Close() error {
	if e.closed { return nil }   // unsynchronized read
	e.closed = true              // unsynchronized write
	return e.pw.CloseWithError(io.EOF)
}
```

`SafeEmulator.Read` deliberately takes **no** lock (so it can block without holding the mutex across the read), and there's no `SafeEmulator.Close` override, so the embedded `Emulator.Close` also runs unlocked (`vt/safe_emulator.go:33`).

The underlying `io.Pipe` is already concurrency-safe — `pw.CloseWithError(io.EOF)` correctly unblocks a blocked `pr.Read`. The **only** unsafe shared state is the `closed` flag.

### Impact

The natural teardown pattern — one goroutine draining `Read()`, another calling `Close()` to unblock it — trips `go test -race` with a data race on `closed` (read at `emulator.go:252`, write at `emulator.go:265`). Consumers are forced to choose between leaking the `Read` goroutine or accepting the race.

### Suggested fix

Make `closed` an `atomic.Bool` (load in `Read`/`Write`, `Swap` in `Close`) — or drop the `closed` short-circuit entirely and rely on the pipe (a closed pipe already makes `pr.Read` return the close error and `pw.Write` return `ErrClosedPipe`). Either way no mutex needs to be held across the blocking read.

Observed on `vt` module version `v0.0.0-20260323091123-df7b1bcffcca`.

toller892:
## Description

Fixes a data race on `Emulator.closed` between `Read()` and `Close()`.

**Problem:** `Emulator.Read` reads `closed` (line 252) then blocks on the pipe. `Emulator.Close` writes `closed = true` (line 265) to unblock it. Both accesses are unsynchronized — `go test -race` flags this as a data race. `SafeEmulator.Read` deliberately holds no lock (so it can block without holding the mutex), and there's no `SafeEmulator.Close` override, so the race also exists through the safe wrapper.

**Fix:** Replace `closed bool` with `sync/atomic.Bool`:
- `Read()`: `closed.Load()` — lock-free check before blocking on the pipe
- `Close()`: `closed.Swap(true)` — atomic check-and-set (returns early if already closed)
- `Write()`: `closed.Load()` — lock-free check before writing

No mutex needs to be held across the blocking read, which is the whole point of `SafeEmulator.Read` not taking the lock.

**Testing:** Added `TestEmulatorCloseDataRace` and `TestSafeEmulatorCloseDataRace` which exercise the concurrent Read+Close pattern with `go test -race`. All existing tests pass.

Fixes #879

### DIFF
--- a/vt/emulator.go
+++ b/vt/emulator.go
@@ -3,6 +3,7 @@ package vt
 import (
 	"image/color"
 	"io"
+	"sync/atomic"
 
 	uv "github.com/charmbracelet/ultraviolet"
 	"github.com/charmbracelet/ultraviolet/screen"
@@ -68,7 +69,7 @@ type Emulator struct {
 	gsingle int // temporarily select GL or GR
 
 	// Indicates if the terminal is closed.
-	closed bool
+	closed atomic.Bool
 
 	// atPhantom indicates if the cursor is out of bounds.
 	// When true, and a character is written, the cursor is moved to the next line.
@@ -249,7 +250,7 @@ func (e *Emulator) Resize(width int, height int) {
 
 // Read reads data from the terminal input buffer.
 func (e *Emulator) Read(p []byte) (n int, err error) {
-	if e.closed {
+	if e.closed.Load() {
 		return 0, io.EOF
 	}
 
@@ -258,17 +259,16 @@ func (e *Emulator) Read(p []byte) (n int, err error) {
 
 // Close closes the terminal.
 func (e *Emulator) Close() error {
-	if e.closed {
+	if e.closed.Swap(true) {
 		return nil
 	}
 
-	e.closed = true
 	return e.pw.CloseWithError(io.EOF) //nolint:wrapcheck
 }
 
 // Write writes data to the terminal output buffer.
 func (e *Emulator) Write(p []byte) (n int, err error) {
-	if e.closed {
+	if e.closed.Load() {
 		return 0, io.ErrClosedPipe
 	}
 
--- a/vt/emulator_close_test.go
+++ b/vt/emulator_close_test.go
@@ -1,0 +1,84 @@
+package vt
+
+import (
+	"io"
+	"sync"
+	"testing"
+)
+
+func TestEmulatorCloseDataRace(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cameronsjo:
Independent confirmation that this is still needed and that the approach holds, from a downstream terminal-multiplexing consumer of `vt`:

- Still reproduces on `v0.0.0-20260730164118-7e2d3e6c5238` (Go 1.26.5, darwin/arm64) — unchanged at `emulator.go:252` (read) against `:265` (write).
- I arrived at the identical `atomic.Bool` change independently before finding this PR, which is a small point in favour of it being the obvious fix. `Swap(true)` here is neater than the `CompareAndSwap` I had.
- Applied this patch locally and ran the downstream project's full suite under `-race`: green with no test exclusions, where previously every test touching emulator shutdown had to be excluded from the race job behind a `!race` build tag.
- `vt`'s own `go test -race ./...` passes with the patch applied.

Worth noting for reviewers wary after #850/#851: this doesn't reintroduce that deadlock risk. #850 added lock coverage around `Read`, which is exactly what deadlocked, and #851 reverted it. This change adds no lock at all — `Read` stays lock-free and the atomic only replaces an early-out flag whose real unblocking is still `pw.CloseWithError(io.EOF)` on an internally-synchronized `io.Pipe`.

Happy to help get this over the line if anything is blocking it.

Issue #848, opened 2026-04-22
ansi: OSC/DCS/SOS/PM/APC strings truncated by UTF-8 continuation byte 0x9C


wblech:
## Summary

The `x/ansi` parser truncates OSC, DCS, SOS, PM, and APC strings whose
payload contains certain UTF-8 multi-byte codepoints, because byte
`0x9C` is both a valid UTF-8 continuation byte and the 8-bit C1 form
of the String Terminator (ST).

When a codepoint whose UTF-8 encoding contains `0x9C` appears inside
one of these string-collecting states, the parser exits the state at
`0x9C` and interprets the remaining payload bytes as ground-state
input. For OSC/DCS/SOS/PM/APC terminated via `ESC \` (7-bit ST), the
payload bytes after `0x9C` end up written into the terminal buffer as
regular characters.

The entire U+2700–U+273F Dingbats block is affected (second UTF-8 byte
is `0x9C`). A concrete example: U+2733 `✳` encodes as `E2 9C B3`, and
is emitted by real-world tools in OSC 0 window-title updates.

## Minimal reproduction

Feed this byte stream through the parser with an OSC 0 handler that
logs the payload:

```
ESC ] 0 ; E2 9C B3 ESC \
```

Expected: handler receives `E2 9C B3` (one `✳`) as the payload; parser
returns to ground after the final `ESC \`.

Actual: parser exits OSC state on `0x9C`; OSC dispatches with payload
`E2` only; bytes `B3 ESC \` are processed as ground input, with `B3`
written to the terminal grid.

## Reproducing tests

The following cases can be dropped into `ansi/parser_osc_test.go` and
`ansi/parser_dcs_test.go` alongside the existing `TestOscSequence` /
`TestDcsSequence` table tests. They fail on current `main` and express
the fixed behaviour: the parser must not let a UTF-8 continuation byte
act as a state-transition byte.

Add to the `cases` slice in `TestOscSequence`
(`ansi/parser_osc_test.go`):

```go
{
    // U+2733 ✳ (E2 9C B3). Real-world OSC 0 window-title input.
    name:  "osc_dingbat_u2733_preserved",
    input: "\x1b]0;✳ hello\x07",
    expected: []any{
        []byte("0;✳ hello"),
    },
},
{
    // U+2736 ✶ (E2 9C B6) — another Dingbat in the 0x9C-second-byte range.
    name:  "osc_dingbat_u2736_preserved",
    input: "\x1b]0;✶ Run\x07",
[...]

PR #886, opened 2026-06-04
fix(ansi/parser): keep UTF-8 string-state payloads intact (drop 8-bit C1 ST)

ktat:
## Problem

The `Anywhere` transition block installs two entries in every state,
including the string states (OSC / DCS / SOS / PM / APC):

- `0x9C` → `ExecuteAction + GroundState` (C1 ST as terminator)
- `0xC2..0xF4` → `CollectAction + Utf8State` (UTF-8 lead bytes)

Both are wrong for UTF-8 string payloads. `0x9C` is a valid UTF-8
continuation byte — U+2733 (✳) is `0xE2 0x9C 0xB3`, U+672B (末) is
`0xE6 0x9C 0xAB` — so treating it as ST splits the string mid-character
and leaks the tail as visible text. The `0xC2..0xF4` entries jump to
`Utf8State` which returns to `GroundState`, abandoning the string state
entirely.

## Fix

Each string state's `AddRange(0x20..0xFF, PutAction, …)` already
overwrites the Anywhere entries — this PR ensures that order is
preserved and removes the explicit `AddOne(0x9C, DispatchAction, …)`
overrides in OSC, DCS, and SOS/PM/APC. For SOS/PM/APC the payload
range is widened from `0x20..0x7F` to `0x20..0xFF`.

The 8-bit C1 ST (`0x9C`) is intentionally dropped as a terminator.
ESC `\` (7-bit ST) and BEL (OSC only) remain supported.

## Tests

New `ansi/parser/transition_table_test.go` verifies directly against
the generated table that `0x9C` and `0xC2..0xF4` are `PutAction` in
all five string states, and that 7-bit terminators still fire.

Existing tests updated to replace `0x9C` terminators with `ESC \`.
Two test cases that relied on 8-bit ST termination are removed.

### DIFF
--- a/ansi/parser/transition_table.go
+++ b/ansi/parser/transition_table.go
@@ -157,14 +157,23 @@ func GenerateTransitionTable() TransitionTable {
 	table.AddOne(']', EscapeState, StartAction, OscStringState)
 
 	// Sos_pm_apc_string
+	//
+	// String payload extends to 0xFF so that the C1 ST (0x9C) and UTF-8 lead
+	// bytes installed by the Anywhere block above are overwritten with
+	// PutAction here — otherwise 0x9C would terminate the sequence (which
+	// collides with the 0x9C continuation byte of multi-byte UTF-8 chars such
+	// as U+2733 ✳ = 0xE2 0x9C 0xB3), and 0xC2..0xF4 would transition into
+	// Utf8State and return to GroundState mid-string. ESC \\ (7-bit ST) and
+	// CAN/SUB are kept as terminators below; the 8-bit C1 ST is intentionally
+	// dropped (callers must use ESC \\).
 	for _, state := range r(SosStringState, ApcStringState) {
 		table.AddRange(0x00, 0x17, state, PutAction, state)
 		table.AddOne(0x19, state, PutAction, state)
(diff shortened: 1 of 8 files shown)


ktat:
Note: I became aware of #848 after opening this PR. The approach proposed there — tracking UTF-8 continuation state so that 0x9C is treated as ST only when it is not part of an active multi-byte sequence — is more correct than this PR, which simply drops 8-bit C1 ST as a terminator entirely.

If a fix along the lines of #848 lands, this PR should be closed in favor of that. Happy to close it now if the maintainers prefer to wait for the more complete fix.

ktat:
Also, the "Known limitation" section in the commit message contains two inaccuracies:

- "xterm.js does this" — xterm.js does not have this problem in the first place because it operates on decoded Unicode codepoints (JavaScript strings are UTF-16),  not raw bytes. There is no UTF-8 reassembly to do.
- "route the ≥0x80 payload bytes through Utf8State" — the approach proposed in #848 is different: keep the state machine as-is and track a UTF-8 continuation byte  counter *outside* it, so that 0x9C is treated as ST only when no multi-byte sequence is in progress.

PR #902, opened 2026-06-18
feat(vt): xterm-style scrollback-aware resize (Scrollback.Pop)

LXXero:
## Problem

`Emulator.Resize` pads/cuts the primary screen at the **bottom** and keeps content **top-anchored**. A repaint-in-place TUI that anchors its UI to the bottom (a shell prompt, Claude Code, anything using cursor-positioned redraws on the primary screen) then repaints over the newly-added bottom space while its **old frame lingers at the top** — a duplicated ghost on every grow until the app is forced to fully redraw. Growing taller also fails to *reveal* scrollback history, which is what users expect when they enlarge a window.

## Change

Make the primary screen resize the way xterm/vte do:

- **Grow taller** pulls up to `delta` lines back **down from scrollback** onto the top, so existing content (and the cursor) stays anchored to the bottom.
- **Shrink** pushes the displaced **top** lines into scrollback (recoverable by scrolling up) and keeps the bottom rows + cursor.

Width changes keep the plain per-line pad/cut (vt/uv never rewrap). The **alt screen** — no scrollback, owned by a full-screen app that repaints on `SIGWINCH` — keeps the plain `Resize`.

## API

Adds `Scrollback.Pop()` — the inverse of `Push` for the newest line (linearizes the wrapped ring on this cold, resize-only path) — and an internal `Screen.resizeReflow`, wired into `Emulator.Resize` for the primary screen.

## Tests

- grow pulls scrollback, content stays bottom-anchored, cursor follows;
- shrink pushes top lines into scrollback, bottom row + cursor preserved;
- `Pop` ordering across a wrapped ring.

Full `vt` suite passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/vt/emulator.go
+++ b/vt/emulator.go
@@ -215,6 +215,16 @@ func (e *Emulator) CursorPosition() uv.Position {
 
 // Resize resizes the terminal.
 func (e *Emulator) Resize(width int, height int) {
+	// Primary screen resizes the xterm way: grow pulls history back down
+	// so content stays bottom-anchored, shrink pushes the displaced top
+	// rows into scrollback. resizeReflow moves scrs[0]'s own cursor row
+	// to follow the content, so the cursor is read back AFTER the resize.
+	// The alt screen has no scrollback and is owned by a full-screen app
+	// that repaints on SIGWINCH, so it keeps the plain pad/cut.
+	e.scrs[0].resizeReflow(width, height)
+	e.scrs[1].Resize(width, height)
+	e.tabstops = uv.DefaultTabStops(width)
+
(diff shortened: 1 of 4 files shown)


taciturnaxolotl:
is there a reason why you closed this?

LXXero:
it didn't seem to actually fix the problems i was having...I have a few other PRs I raised that have measurable performance impact, so I didn't want to waste anyone's time with something I couldn't prove effective. If this is something you find valuable, I could re-open it.

taciturnaxolotl:
ah alright!

PR #931, opened 2026-07-30
vt: make Emulator.closed atomic to fix the Close/Read data race

### LINKED ISSUE #930: vt: data race on Emulator.closed between Close and a parked Read
## Summary

`vt.Emulator.closed` is a plain `bool` that `Close` writes and `Read` reads with no synchronization between them, so `-race` flags every shutdown of an `Emulator` whose output is being consumed.

This is not avoidable by using the type carefully. `Read` blocks on an empty pipe, and `Close` is the only thing that unblocks it, so "one goroutine reads, another closes" is the only shape a correct shutdown can take.

## Reproducer

Fails reliably under `go test -race` on `vt v0.0.0-20260730164118-7e2d3e6c5238`:

```go
package vtrace

import (
	"testing"
	"time"

	"github.com/charmbracelet/x/vt"
)

func TestCloseRacesParkedRead(t *testing.T) {
	e := vt.NewEmulator(80, 24)

	done := make(chan struct{})
	go func() {
		defer close(done)
		buf := make([]byte, 64)
		for {
			if _, err := e.Read(buf); err != nil {
				return
			}
		}
	}()

	time.Sleep(100 * time.Millisecond) // let the reader park inside Read
	if err := e.Close(); err != nil {
		t.Fatalf("Close: %v", err)
	}
	<-done
}
```

```
WARNING: DATA RACE
Read at 0x00c000112160 by goroutine 8:
  github.com/charmbracelet/x/vt.(*Emulator).Read()
      vt/emulator.go:252

Previous write at 0x00c000112160 by goroutine 7:
  github.com/charmbracelet/x/vt.(*Emulator).Close()
      vt/emulator.go:265
```

## Detail

`vt/emulator.go`:

```go
// Read reads data from the terminal input buffer.
func (e *Emulator) Read(p []byte) (n int, err error) {
	if e.closed {          // :252 — read
		return 0, io.EOF
	}
	return e.pr.Read(p)
}

// Close closes the terminal.
func (e *Emulator) Close() error {
	if e.closed {          // :261 — read
		return nil
	}
	e.closed = true        // :265 — write
	return e.pw.CloseWithError(io.EOF)
}
```

`SafeEmulator` does not help here, which may be worth noting given its doc comment says it "adds concurrency safety":

- its `Read` deliberately takes no lock (`vt/safe_emulator.go:33`), which is presumably intentional so a parked reader can't hold `mu` against every writer, and
- it does not override `Close` at all, so the embedded `*Emulator.Close` is promoted unchanged.

So the one pair that actually races is the one pair the wrapper doesn't cover.

## Suggested fix

The guard is redundant rather than load-bearing: the real unblocking is `e.pw.CloseWithError(io.EOF)` on an `io.Pipe`, which is internally synchronized and already idempotent. `closed` is only an early-out, so it doesn't need a mutex — making it an `atomic.Bool` is enough, and keeps `Read` lock-free:

```go
closed atomic.Bool
```

with `e.closed` → `e.closed.Load()` at `:252`, `:261` and `:271`, and `e.closed = true` → `e.closed.Store(true)` at `:265`.

Happy to send that as a PR if the approach looks right.

## Impact

Downstream this forces a choice between dropping `-race` from CI and excluding every test that exercises emulator shutdown, since there is no usage pattern that avoids it.

cameronsjo:
Closes #930

`Emulator.closed` is a plain `bool` that `Close` writes and `Read` reads with no synchronization between them, so `-race` flags every shutdown of an `Emulator` whose output is being consumed. `Read` blocks on an empty pipe and `Close` is the only thing that unblocks it, so there is no usage pattern that avoids the pair running concurrently.

`SafeEmulator` does not cover it either: its `Read` takes no lock (deliberately, so a parked reader can't hold `mu` against every writer) and it does not override `Close`, so the embedded `*Emulator.Close` is promoted unchanged.

## The change

`closed` becomes an `atomic.Bool`. The guard is an early-out rather than the mechanism — the actual unblocking is `pw.CloseWithError(io.EOF)` on an `io.Pipe`, which is internally synchronized and idempotent — so no mutex is needed and `Read` stays lock-free.

`Close` uses `CompareAndSwap` rather than a `Load` then a `Store`, which also makes it single-winner: two concurrent `Close` calls previously could both pass the guard and both call `CloseWithError`. Harmless, since that call is idempotent, but the CAS form costs nothing and says what it means.

## Verification

- `go test -race ./...` in `vt/` passes.
- A reproducer that reliably fails on `main` passes with this change:

```go
func TestCloseRacesParkedRead(t *testing.T) {
	e := vt.NewEmulator(80, 24)

	done := make(chan struct{})
	go func() {
		defer close(done)
		buf := make([]byte, 64)
		for {
			if _, err := e.Read(buf); err != nil {
				return
			}
		}
	}()

	time.Sleep(100 * time.Millisecond) // let the reader park inside Read
	if err := e.Close(); err != nil {
		t.Fatalf("Close: %v", err)
	}
	<-done
}
```

On `main` this reports the race at `emulator.go:252` (Read) against `emulator.go:265` (Close). With this change it passes.

- Downstream check: a terminal-multiplexing consumer of `vt` that had to exclude every shutdown-touching test from its `-race` job runs its full suite green, with no exclusions, against a patched `vt`.

I did not add the reproducer as a test in this PR since it needs a sleep to set up the condition; happy to include it in whatever form suits the package if you'd like it covered.

### DIFF
--- a/vt/emulator.go
+++ b/vt/emulator.go
@@ -3,6 +3,7 @@ package vt
 import (
 	"image/color"
 	"io"
+	"sync/atomic"
 
 	uv "github.com/charmbracelet/ultraviolet"
 	"github.com/charmbracelet/ultraviolet/screen"
@@ -67,8 +68,9 @@ type Emulator struct {
 	gl, gr  int
 	gsingle int // temporarily select GL or GR
 
-	// Indicates if the terminal is closed.
-	closed bool
+	// Indicates if the terminal is closed. Atomic because Close is the only
+	// way to unblock a parked Read, so the two necessarily run concurrently.
+	closed atomic.Bool
 
 	// atPhantom indicates if the cursor is out of bounds.
 	// When true, and a character is written, the cursor is moved to the next line.
@@ -249,7 +251,7 @@ func (e *Emulator) Resize(width int, height int) {
 
 // Read reads data from the terminal input buffer.
 func (e *Emulator) Read(p []byte) (n int, err error) {
-	if e.closed {
+	if e.closed.Load() {
 		return 0, io.EOF
 	}
 
@@ -258,17 +260,16 @@ func (e *Emulator) Read(p []byte) (n int, err error) {
 
 // Close closes the terminal.
 func (e *Emulator) Close() error {
-	if e.closed {
+	if !e.closed.CompareAndSwap(false, true) {
 		return nil
 	}
 
-	e.closed = true
 	return e.pw.CloseWithError(io.EOF) //nolint:wrapcheck
 }
 
 // Write writes data to the terminal output buffer.
 func (e *Emulator) Write(p []byte) (n int, err error) {
-	if e.closed {
+	if e.closed.Load() {
 		return 0, io.ErrClosedPipe
 	}
 


cameronsjo:
Closing this — #881 already proposes the same `atomic.Bool` change and does it better, using `Swap(true)` for the close-once guard and adding a regression test this PR lacks. I missed it before opening; apologies for the noise.

Moving the verification detail over to #881.

